PR #16, opened 2020-12-11
Fixed walls (so no more, or at least less, I'm not exactly sure, log spam) and Bronze Dungeon Structure Files

reetamb:


reetamb:
mcmeta is fixed Wow

PR #27, opened 2021-01-25
[WIP] Added a common and client config file to the mod. Added the zephyr and aerwhale legacy models.

Burning-Cactus:
This pull request adds configuration files to the mod. I added options for everything from the 1.12 version except for the dimension ID. Currently, none of the config options do anything except for the legacy models option, which now toggles between the new and old models for zephyrs and aerwhales.
There is also a custom config GUI registered. It's very bare-bones and unfinished, but it should be convenient for swapping config values during testing.


Burning-Cactus:
Closed this pull request as it is now obsolete.

PR #98, opened 2021-04-03
animated aether portal texture

trolls-incorporated:
current aether portal texture looks a bit ugly compared to the nether portal texture, this one looks more like the nether portal texture and is less ugly


bconlon1:
Textures are not being redone at this time. The programmer art textures are staying how they are.

PR #133, opened 2021-05-12
Small menu fixes

StonedCoal:
I added the mod to my 1.16 Modpack and noticed some bugs:

- Getting the splash via reflection doesn't work due to obfuscation. So we now just get a new splash.
- Reordering the buttons with a seperate button counter allows for some degree of mod compatibility.
  e.g. quark adds a "q"-Button which is now displayed under the "Quit Game"-Button


bconlon1:
Added these changes to the recent commit.

PR #147, opened 2021-06-12
Falling out of the Aether

Burning-Cactus:
With this PR, entities can get to the Overworld by falling below y=0 in the Aether. In addition, any of the entity's passengers and their derivative passengers will be transported along with them and continue riding.


Burning-Cactus:
Oh, and a couple of minor optimizations regarding the Sky Renderer and mod setup.

PR #159, opened 2021-07-17
aether menu screen now renders the last world you played on

Kelvin285:
**edit: there's one last bug that needs to be fixed.  You stay in spectator mode after joining a world.  Fixing it right now**
added a change to the main menu panorama where the main menu uses the last world you played in as the panorama.  Mobs spawn, the sun and moon move, etc.
image
image


Kelvin285:
closed because of the spectator mode bug.  I'll re-upload it once that's fixed

PR #295, opened 2021-12-20
add swet shrink animation

baguchi:
I add swet shrink animation like aether 2
when swet is shrink It also weakens the power to launch the player


bconlon1:
Port this to 1.18 so it can be reviewed.

PR #304, opened 2021-12-29
[Draft - Just for review] 1.18 port reflection

Jaryt:
Opening this PR so we can see the changes that have been made during the 1.18 port and double check our work. If there is anything that seems off, please comment at that line so we can be confident any bad changes won't crop up in the future


src/generated/resources/data/aether/dimension/the_aether.json
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "aether:aether_type",
+  "generator": {
+    "biome_source": {
+      "biome": "aether:floating_forest",
```
bconlon1, line 5: Biome should probably just be renamed to the_aether.

src/main/java/com/gildedgames/aether/Aether.java
```diff
@@ -73,6 +83,7 @@ public Aether() {
 		modEventBus.addListener(this::commonSetup);
 		modEventBus.addListener(this::curiosSetup);
 		modEventBus.addListener(this::dataSetup);
+		modEventBus.addListener(AetherCapabilities::register);
```
bconlon1: An event subscriber annotation can probably be added to AetherCapabilities instead of doing it through here.

src/main/java/com/gildedgames/aether/client/AetherClient.java
```diff
@@ -26,13 +29,12 @@
     public static void clientInitialization() {
         IEventBus modEventBus = FMLJavaModLoadingContext.get().getModEventBus();
         modEventBus.addListener(AetherClient::clientSetup);
-        modEventBus.addListener(AetherClient::clientComplete);
+        modEventBus.addListener(AetherRendering::registerEntityRenderers);
```
bconlon1: Might be able to use an event subscriber annotation here instead, not sure if that'd be best here though.

bconlon1: Maybe could be worth it to just do that and scrap clientInitialization(), and have the call to it call to clientSetup() instead (or well, register it instead since it's an event), assuming this doesn't break anything from it being client only and registered from non-client only code.

src/generated/resources/data/aether/dimension/the_aether.json
bconlon1: Or skylands or something, since the_aether is the dimension name that may be weird.

PR #305, opened 2021-12-30
refactor: Changes to some client classes and switched to use annotations for even subscribers in some places.

bconlon1:
Renamed AetherRendering to AetherRenderers, and moved anything in it that wasn't related to a renderer into AetherClient. Also added annotations for some event subscribers instead of registering through the event bus methods.


Jaryt (approved):
lgtm

PR #309, opened 2021-12-30
RenderFixed: Sheepuff and Skyroot Bed

baguchi:
I fixed Sheepuff model and skyroot bed item render
The skyroot bed is not very good code so please let's me know if you have any problems


src/main/java/com/gildedgames/aether/client/renderer/entity/model/SheepuffModel.java
```diff
@@ -30,6 +33,7 @@ public void prepareMobModel(SheepuffEntity sheepuff, float limbSwing, float limb
         super.prepareMobModel(sheepuff, limbSwing, limbSwingAmount, partialTick);
         this.head.y = 6.0F + sheepuff.getHeadRotationPointY(partialTick) * 9.0F;
         this.headXRot = sheepuff.getHeadEatAngleScale(partialTick);
+        this.body.xRot = ((float) Math.PI / 2F);
```
Jaryt, line 32: Hi Bagu! Is this to fix the extra body layer that is sticking out of the back? If so, we can discard this change since we'll probably be solving that in a different way :)

Jaryt (approved):

PR #327, opened 2022-01-01
refactor: Refactor our structures from Jigsaw

Drullkus:
We'd be better off using extensions of `TemplateStructurePiece` to get the control we need over our structure pieces including having critical data-driven customization. This refactor will accomplish this while restoring functional parity with Aether b1.7.3's generation methods.


bconlon1:
No longer relevant, see: #811.

PR #343, opened 2022-01-03
refactor: Simplified BEWLR registering.

bconlon1:
Simplified how we register BEWLRs. This should ideally mean we don't have to change the AetherBlockEntityWithoutLevelRenderer every time we add a new BEWLR for an item, we can just register in the same place as the block item itself.

### DIFF
--- a/src/main/java/com/gildedgames/aether/common/item/block/EntityBlockItem.java
+++ b/src/main/java/com/gildedgames/aether/common/item/block/EntityBlockItem.java
@@ -1,22 +1,23 @@
 package com.gildedgames.aether.common.item.block;
 
-import com.gildedgames.aether.common.registry.AetherBlocks;
-import com.mojang.blaze3d.vertex.PoseStack;
-import net.minecraft.client.Minecraft;
 import net.minecraft.client.renderer.BlockEntityWithoutLevelRenderer;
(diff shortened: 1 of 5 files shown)


src/main/java/com/gildedgames/aether/common/item/block/EntityBlockItem.java
```diff
@@ -1,30 +1,31 @@
...
-public class SkyrootBedItem extends BlockItem {
-    public <B extends Block> SkyrootBedItem(B block, Properties tab) {
+public class EntityBlockItem extends BlockItem {
+    private Supplier<BlockEntityWithoutLevelRenderer> renderer;
+
+    public <B extends Block> EntityBlockItem(B block, Properties tab) {
         super(block, tab);
     }
 
+    public EntityBlockItem setBEWLR(Supplier<BlockEntityWithoutLevelRenderer> renderer) {
```
bconlon1, line 18: setBWLR could potentially be used to add to a map in BlockEntityWithoutLevelRenderer instead of creating a whole new one for every block entity.

bconlon1: This may require reverting the changes that removed AetherBlockEntityWithoutLevelRenderer's initialization in AetherBlocks.

PR #344, opened 2022-01-04
refactor: Another iteration of BEWLR refactoring.

bconlon1:
Simplified how we register BEWLRs. This should ideally mean we don't have to change the AetherBlockEntityWithoutLevelRenderer every time we add a new BEWLR for an item, we can just register in the same place as the block item itself by passing a block entity as an argument.

### DIFF
--- a/src/main/java/com/gildedgames/aether/client/renderer/tile/AetherBlockEntityWithoutLevelRenderer.java
+++ b/src/main/java/com/gildedgames/aether/client/renderer/tile/AetherBlockEntityWithoutLevelRenderer.java
@@ -1,6 +1,5 @@
 package com.gildedgames.aether.client.renderer.tile;
 
-import com.gildedgames.aether.common.entity.tile.SkyrootBedBlockEntity;
 import com.gildedgames.aether.common.item.block.EntityBlockItem;
 import com.gildedgames.aether.common.registry.AetherBlocks;
 import com.mojang.blaze3d.vertex.PoseStack;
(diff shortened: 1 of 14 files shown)


src/main/java/com/gildedgames/aether/client/renderer/tile/AetherBlockEntityWithoutLevelRenderer.java
```diff
@@ -12,28 +11,36 @@
...
-
         Item item = pStack.getItem();
-        if (item instanceof BlockItem blockItem) {
-            if (blockItem.getBlock() instanceof SkyrootBedBlock) {
-                Minecraft.getInstance().getBlockEntityRenderDispatcher().renderItem(skyrootBedEntity, pPoseStack, pBuffer, pPackedLight, pPackedOverlay);
+        if (item instanceof EntityBlockItem blockItem) {
+            Block block = blockItem.getBlock();
+            BlockEntity blockEntity = null;
+            if (block == AetherBlocks.SKYROOT_BED.get()) {
+                blockEntity = this.skyrootBed.get();
```
Drullkus: Calling `skyrootBed.get()` is going to initialize a new `SkyrootBedBlockEntity` each render frame

Drullkus (changes requested):
It mostly looks good to me, just one small change required

Drullkus (approved):
Looks beautiful

Jaryt (approved):
wonderful!

PR #347, opened 2022-01-04
refactor: Breaking Change: Rewrote freezing behavior of blocks and items

Drullkus:
This addresses Icestone bringing server TPS to its knees


src/main/java/com/gildedgames/aether/common/block/construction/IcestoneSlabBlock.java
```diff
@@ -1,37 +1,30 @@
...
+        worldIn.scheduleTick(pos, this, 10);
     }
 
     @Override
     public void tick(BlockState state, ServerLevel worldIn, BlockPos pos, Random random) {
         super.tick(state, worldIn, pos, random);
-        freezeFluids(worldIn, pos);
-        ScheduledTick<Block> scheduledTick = new ScheduledTick<>(this, pos, 10, 0);
-        worldIn.getBlockTicks().schedule(scheduledTick);
+        this.freezeBlocks(worldIn, pos, state, 2.82842712475f); // Square root of 8
```
bconlon1: This is nitpicky but couldn't you just write Math.sqrt(8)?

Drullkus: I'm not really a fan of wrapping a constant with a function so I opted to put the comment to explain the otherwise "magic" number instead

src/main/java/com/gildedgames/aether/common/block/util/FreezingBehavior.java
```diff
@@ -0,0 +1,94 @@
...
+import net.minecraft.tags.FluidTags;
+import net.minecraft.sounds.SoundSource;
+import net.minecraft.sounds.SoundEvents;
+import net.minecraft.core.BlockPos;
+import net.minecraft.world.level.Level;
+
+import java.util.Map;
+import java.util.function.Supplier;
+
+public interface FreezingBehavior<T>
```
bconlon1, line 18: Would personally prefix an interface class with "I" as thats been the standard in the repo.

bconlon1 (approved):
lgtm

PR #351, opened 2022-01-04
fix: Updated elevation check for Aether Portals

### LINKED ISSUE #345: Aether Dimension not generating on a superflat world
The portal insta breaks whenever you try to light it up, using the portal frame item does not work as well
It also gives me this weird error in my logs
image

Drullkus:
Fixes #345


Burning-Cactus (approved):
Looks good!

PR #355, opened 2022-01-06
Add en_UD.json (Upside down English translation)

ZolaKluke:
This PR contains a full translation of everything (besides lore since it breaks the book format) to "upside down English"


bconlon1 (approved):
lgtm

PR #356, opened 2022-01-06
Fix: Fixed a crash on the server caused by CapeEntity.

Burning-Cactus:
`OnlyIn`  annotations are pain.

Added a level.isClientSide check before checking for the ClientWorld class.


Burning-Cactus:
Oh my god the annotation tag mentioned an actual person what the heck

bconlon1 (approved):
lgtm

PR #368, opened 2022-01-09
feat: Implement Voronoi Chunk Generator

Drullkus:
- [x] Implement Gridded Generator
- [x] Implement Vectored Generator
- [ ] Implement Voronoi Generator

PR #374, opened 2022-01-10
improved generation

Ultrasquid9:
the generation now has hills and more height variation. pictures were posted in #general-aether on the discord server.

### DIFF
--- a/src/generated/resources/data/aether/worldgen/noise_settings/skyland_generation.json
+++ b/src/generated/resources/data/aether/worldgen/noise_settings/skyland_generation.json
@@ -86,23 +86,23 @@
     "terrain_shaper": {
       "offset": 0.0,
       "factor": 0.0,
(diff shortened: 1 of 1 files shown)


Drullkus (changes requested):
I appreciate your helping out! Unfortunately there's a few things

- You modified a data-generated file directly, these changes will not persist unless you write these changes to `AetherWorldProvider.java`'s `aetherNoiseSettings()`
- #368 is currently in progress which may erase these changes because I'm implementing a custom Chunk Generator instead of using the regular Noise Chunk Generator

I'd hold off from working on this PR further until #368 is completed

bconlon1:
Considering this PR never got fixed with the requested changes and is also outdated at this point, I am going to close it. Either way, I don't see much point in implementing this when it would likely be overwritten completely by our inevitable rework of the world generation.

PR #357, opened 2022-01-06
fix: Fixed Curios models, renderers, and also other Curios related issues that appeared after the port to 1.18.

bconlon1:
- [x] #254 
- [x] #247
- [x] #350 
- [x] #259 
- [x] #258
- [x] #152
- [x] #363
- [x] #383 
- [x] #228

PR #387, opened 2022-01-12
feat: Ported the content from feature/rankings_system.

bconlon1:
This includes the Player Halo and Developer Glow, the rankings system rework, as well as the WIP Customization & Perks screen accessible from the Accessories menu.

This is probably going to conflict in some way with #357.


Jaryt (approved):
lgtm!

We'll probably offload the UUID storage into the patreon access system in the future for better privacy/ease of sync

PR #396, opened 2022-01-15
Fix: The world tick listener now checks against the minimum build height.

### LINKED ISSUE #394: Improve height check determining if a player has fallen out of the aether
Currently it checks against `0` instead of `world.getMinBuildHeight()`. This isn't a problem because we haven't deviated our world heights but this will be if a datapacker does

image

I also recommend not counting the player as falling if either: Player's `Abilities.flying == true` or `player.isSpectator()`

Burning-Cactus:
Players now fall out of the Aether at the dimension type's minimum build height instead of y=0. Players flying through creative or spectator can't fall out of the Aether.
Closes #394

### DIFF
--- a/src/main/java/com/gildedgames/aether/common/event/listeners/DimensionListener.java
+++ b/src/main/java/com/gildedgames/aether/common/event/listeners/DimensionListener.java
@@ -146,7 +146,7 @@ public class DimensionListener
                     if (!AetherConfig.COMMON.disable_falling_to_overworld.get()) {
                         for (Entity entity : world.getEntities(EntityTypeTest.forClass(Entity.class), Objects::nonNull)) {
                             if (entity.getY() <= world.getMinBuildHeight() && !entity.isPassenger()) {
-                                if(!(entity instanceof Player player && player.getAbilities().flying)) {
+                                if (!(entity instanceof Player player && player.getAbilities().flying)) {
                                     fallFromAether(entity);
                                 }
                             }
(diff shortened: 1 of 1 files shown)


src/main/java/com/gildedgames/aether/common/event/listeners/DimensionListener.java
```diff
@@ -145,7 +145,10 @@ public static void onWorldTick(TickEvent.WorldTickEvent event) {
                 if (event.phase == TickEvent.Phase.END) {
                     if (!AetherConfig.COMMON.disable_falling_to_overworld.get()) {
                         for (Entity entity : world.getEntities(EntityTypeTest.forClass(Entity.class), Objects::nonNull)) {
-                            if (entity.getY() <= 0 && !entity.isPassenger()) {
+                            if (entity.getY() <= world.getMinBuildHeight() && !entity.isPassenger()) {
+                                if(entity instanceof Player player && (player.getAbilities().flying)) {
```
Drullkus: Please also check if the player is Spectator

Drullkus (changes requested):

src/main/java/com/gildedgames/aether/common/event/listeners/DimensionListener.java
Drullkus: Spectator is also considered flying, resolved

Drullkus (approved):

Drullkus (approved):
Looks good to me!

bconlon1 (approved):
lgtm

PR #435, opened 2022-01-20
Updated saving data methods for some block entities.

Burning-Cactus:
`BlockEntity.save(CompoundNBT)` no longer exists. TreasureChestBlockEntity and IncubatorTileEntity now override saveAdditional instead, ensuring that the mod is compatible with the latest Forge version.


bconlon1 (approved):
lgtm

PR #440, opened 2022-01-23
Feat: The Aether fog rendering applies to all Aether dimensions now.

Burning-Cactus:
Any dimension that uses an instance of AetherSkyRenderInfo for its effects won't get black fog near the minimum y value of the world. If the minimum y value is below 0, the fog color event handler now accounts for that.


bconlon1 (approved):
lgtm

PR #454, opened 2022-02-08
create ja_jp translate

baguchi:
I made ja_jp translate
I was able to translate only part of it, but I hope you'll be happy.


bconlon1 (approved):
Might as well approve this now; if there's any issues with it, they can be handled by OneSky contributors in the future when we use that for our translations.

PR #469, opened 2022-03-01
refactor: Bed behavior changes.

### LINKED ISSUE #465: Refactor: Remove Skyroot Bed explosiveness in the Overworld.
While technically a vanilla-consistent feature, it's likely to catch player's off guard and be more annoying than fun to deal with.

Maybe they should also be allowed to be slept in when in the Overworld, although alternatively they can just be completely useless.

Should vanilla beds no longer explode in the Aether as well because of this?

bconlon1:
- [x] Closes #465


Burning-Cactus (approved):
Looks good to me!

PR #470, opened 2022-03-01
refactor: Various refactors.

### LINKED ISSUE #187: Bug: EntityDataManager does not like me storing data variables in the player capability.
Look at:
- SynchedEntityData
- ServerEntity.sendDirtyEntityData()
- ClientboundSetEntityDataPacket

Create a class `SynchedCapabilityData` which will have similar function to `SynchedEntityData` but its purpose will be usage in our capabilities to allow for data syncing without creating tons of packets.

Ideas:
- Might be possible to have a CapabilityDataEntry class which can store basic values like ints and booleans which can be passed into a catch-all packet to send that data across sides to sync.

### LINKED ISSUE #281: Feature: Debugging commands
- Should allow for resetting certain capability data in the event of mod incompatibilities breaking things, e.g. life shards.
- Need to make sure they get run properly on client and server for capability values to avoid desync.
- Organize `command` package.

### LINKED ISSUE #390: Refactor: Move code from being stored directly in event listeners to methods that can be called from event listeners.
- Anything that uses an event as a parameter should probably be switched to instead use values from the event as parameters where necessary. E.g. ZaniteAccessory.

### LINKED ISSUE #441: Chore: Use Item.interactLivingEntity(ItemStack, Player, LivingEntity, InteractionHand) instead of EntityInteractSpecific event

### LINKED ISSUE #448: Chore: Use .resetFallDistance() to cancel fall distance.

### LINKED ISSUE #467: Chore: Move more jsons over to DataGen
- Curios tags if possible.
- The remaining blockstates and block models (may require custom methods).

bconlon1:
- [x] Will include all the changes from #443.
- [x] Closes #467 
- [x] Closes #448


Burning-Cactus (approved):
Looks good to me!

PR #486, opened 2022-03-13
fix: Increased compatibility for dispenser behavior by not completely overriding behavior for fire charges and flint and steel.

### LINKED ISSUE #458: Enhancement: Use mixin instead of overriding dispenser behaviors.
We probably should not be overriding certain dispenser behaviors to prevent them from working in the Aether for item banning like with fireballs. Mixins seem like the best solution if there aren't any other solutions.

bconlon1:
- [x] Closes #458.


Burning-Cactus (approved):
Looks good to me!

PR #490, opened 2022-03-14
feat: Sun Altar player whitelist system.

### LINKED ISSUE #455: Feature: Whitelist for Sun Altar usage.
Have a UUID whitelist file that determines what players can use the Sun Altar on servers.

bconlon1:
- Changed config for the Sun Altar to determine that whitelist.
- Moved some stuff out of the server package.
- Closes #455.


Burning-Cactus (changes requested):
The execute method in SunAltarPacket is wrong. If the sun altar whitelist is disabled, then the sun altar is unusable, and if the player is an op, the sun altar is unusable.

bconlon1:
Forgot to invert the check in the packet, I'll probably only have to replace it with
`if (player != null && (!AetherConfig.COMMON.sun_altar_whitelist.get() || player.hasPermissions(4) || SunAltarWhitelist.INSTANCE.isWhiteListed(player.getGameProfile())))`

Burning-Cactus (approved):
Looks good to me!

PR #493, opened 2022-03-14
fix: Whirlwind now plays item drop sound in the correct code location.

### LINKED ISSUE #492: Bug: Whirlwind drop sound plays even when not dropping items.
https://github.com/Gilded-Games/The-Aether/blob/e74f4000b79ca0b3ee19f96850f6815a1cdfce57/src/main/java/com/gildedgames/aether/common/entity/monster/AbstractWhirlwind.java#L116 should be moved to be next to
https://github.com/Gilded-Games/The-Aether/blob/e74f4000b79ca0b3ee19f96850f6815a1cdfce57/src/main/java/com/gildedgames/aether/common/entity/monster/AbstractWhirlwind.java#L177

bconlon1:
- [x] Closes #492.


Burning-Cactus (approved):
Looks good to me!

PR #494, opened 2022-03-17
Fix: Aerclouds are now valid spawn blocks.

### LINKED ISSUE #498: Chore: Remove blue aerclouds from being a zephyr spawning block.

Burning-Cactus:
- [x] Zephyrs can now spawn on aerclouds.
- [x] Closes #498


bconlon1 (changes requested):
Aercloud's acting as a valid spawn block should probably only be true if the mob is in a tag for entities spawnable on aerclouds and not be based on light level (leaves do something similar with ocelots, shouldnt be too hard to implement). This would make cockatrices not need aerclouds on their blacklist and prevent any other unexpected mobs from spawning on them.

bconlon1 (approved):
lgtm

PR #500, opened 2022-03-17
refactor: Make Crystal Tree into Tree Feature

reetamb:
Custom foliage placer generates the Crystal Tree leaves. Island feature runs .place() with the Tree Feature on top of its center block. Holiday tree was started, but NOT FINISHED, so it doesn't work.


bconlon1:
Likely not going to review/merge until the 1.18.2 porting branch is in a better state (so when this is merged with that branch as well, it'll be easy to make sure nothing broke from it that previously worked).

bconlon1 (approved):
lgtm

PR #505, opened 2022-03-18
feat: Berry Bushes can now be silk touched.

### LINKED ISSUE #503: Enhancement: Allow obtaining Blueberry Bushes through silk touch.
Right now breaking the bush always converts back to the stem, however it could be good to allow the player to obtain the bush in full if silk touch is used. This will require some figuring out though due to how the destroy() method used for replacing the bush with a stem has no player context.

bconlon1:
- [x] Closes #503.


Drullkus (approved):
Niceeee!

PR #506, opened 2022-03-19
Aether Dirt path side 2 pixel fix

Mr-Aerod:
Fixes 2 pixels on dirt part of the Aether path texture so it is the same as regular path.
More info: https://discord.com/channels/118816101936267265/569972352373358653/9546598425245819


bconlon1 (approved):
lgtm

PR #521, opened 2022-03-29
refactor: Item ability refactors and fixes.

### LINKED ISSUE #390: Refactor: Move code from being stored directly in event listeners to methods that can be called from event listeners.

### LINKED ISSUE #409: Bug: Skyroot axes don't double block drops

Block, Item

### Other Type

_No response_

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

39.0.40

### The Aether 1 Version

1.0.0

_No response_

### Client Log

https://gist.github.com/kodadacod/a0ebcc81293d2ddbc5d7e517a0b7e84f

_No response_

### Steps to Reproduce

1. Obtain a Skyroot Axe
2. Find a Skyroot Log (with the doubledrops tag)
3. Break the log with the Skyroot Axe

### What You Expect To Happen

The Skyroot Log would drop 2 Logs

### What Actually Happened

The Skyroot Log dropped one Log

### Additional Details

_No response_

### LINKED ISSUE #457: Enhancement: Blacklist for blocks that Gravitite Tools' ability works on.
Probably a good idea for compatibility in case unforeseen problems arise with using Gravitite Tools on other mods' blocks.

### LINKED ISSUE #472: Enhancement: Use events and tool tags for tool abilities instead of interface classes.
We already do this somewhat with the DoubleDrops loot function using SKYROOT_TOOLS (although that could probably be abstracted somewhat into the loot table datagen, I'm not sure if we even need a custom loot function to do that anymore).

### LINKED ISSUE #520: Enhancement: Tags for weapons with abilities for usage in events.
e.g. HOLYSTONE_WEAPON

### LINKED ISSUE #523: Chore: Verify the math for the Zanite accessories are correct.

### LINKED ISSUE #537: Bug: Zanite Sword ability disabled with Custom Attributes
**Minecraft Version: 1.18.2
Forge Version: 40.0.36
Aether Version: 1.0.0**

**What Happened:** 

On the ANTS, I exhausted the durability of two Zanite Swords by testing them on Zombies. One of them was a regular sword as coded, while the other was a custom sword with modified attributes, specifically a power level of 4 (the default for a regular Zanite Sword) and a speed level of 3.2 (When customizing one attribute of a sword, you have to customize all of them, otherwise the attribute[s] in question will be totally absent)

Testing showed that the Regular Sword increased in power over time. In the high range of durability, the Zanite Sword, at regular speed, took 5 hits to kill a Zombie. At mid range, it took 3-4 depending on critical hits. At low durability, it was a solid, consistent 3.

With the modified sword however, at all ranges of durability, it took an entirely consistent 4 Hits per Zombie. This has led me to conclude based on the extensive testing that a Zanite Sword's power increments are disabled when spawned in with customized attributes. 

I believe that this issue may only be exclusive to the Zanite Sword or weapons that add a bonus to attack power; when I modified a Skyroot Sword in the same way, the Double Loot Bonus was still intact, as killing Sheep always resulted in Mutton/Wool drops in multiples of 2 with both regular and modified Skyroot Swords. As for writing this, I have yet to test this with any of the other Aether Weapons.

Thank you for your attention to this issue.

bconlon1:
- [x] Closes #390.
- [x] Closes #523.
- [x] Closes #537.
- [x] Closes #457.
- [x] Closes #472.
- [x] Closes #520.
- [x] Closes #409.


Burning-Cactus (approved):
Looks good to me!

Issue #460, opened 2022-02-24
Bug: Server Crash


WenXin20:
### What Feature Types Apply to This Bug?

Server

### Forge Version

39.0.79

### The Aether 1 Version

1.0.0

### Client Log

n/a

### Crash Report (if applicable)

https://gist.github.com/WenXin20/fabb54aa062b9a6d35e7a8f31d05abd4

### Steps to Reproduce

1. Load mod on a server
2. Crash

### What You Expect To Happen

No server crashes.

### What Actually Happened

The server crashed.

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.

bconlon1:
Does this crash occur with only Aether and its dependencies installed and no other mods? It's odd that a minecraft class is not being found here.

WenXin20:
Ooh, I didn't realize there were dependencies besides forge.  It might take a bit before I can test without other mods. 
Maybe its because the models are only loaded client-side and not on the server?

bconlon1:
Incomplete issue, closing now that it's been almost a month.

bconlon1:
I believe another mod was exposing this issue somehow, as this issue seems to have arisen now for Aether on its own with 1.18.2 related changes despite nothing being done to the class that was crashing. An actual fix for this is in progress.

PR #524, opened 2022-03-30
fix: Moved render properties out of EntityBlockItem so it's not loaded server-side.

bconlon1:
This fixes the server crash issue so ANTS can load and is also technically a resolution for #460.


Burning-Cactus (approved):
Looks good to me!

PR #532, opened 2022-04-02
fix: Fix server crash from entering the Aether.

### LINKED ISSUE #525: Chore: Remove init() for tags.
With the switch to TagKeys, this is no longer necessary.

### LINKED ISSUE #531: Bug: Dimension Tags do not properly function on servers.
**UPDATE AS OF 4/2/2022:**
- This issue is, as far as we are concerned, a vanilla issue. 
- `Level#dimensionTypeRegistration` is one way of accessing a `Holder<DimensionType>` for comparing to a tag, however when called from the side of a client on a multiplayer server, this holder is a _Direct_ holder, which means it stores no tags or `ResourceKey`s, only the `DimensionType` object. In any other situation it is a _Reference_ holder which is capable of being compared to a tag just fine.
- Another way of trying to get a `Holder<DimensionType>` would involve trying to use `Level#registryAccess`, however once again the client-side on a multiplayer server has some issues with this, as the `DimensionType` object that can be pulled from `Level#dimensionType` does not seem to exist in the registry access according to any method that takes `DimensionType` as a parameter; they all return null or some other value meant to represent null even when in any other situation like in singleplayer they return values like the `ResourceKey` of the dimension type just fine.

bconlon1:
This PR removes DimensionType tags to fix the issue, see: #531.
- [x] Closes #525.


Burning-Cactus (approved):
Looks good to me!

PR #538, opened 2022-04-03
Enhancement: Switch the incubator to use recipes instead of hardcoding it.

### LINKED ISSUE #529: Enhancement: Convert Incubator behavior to be recipe-based.
Would really only just act as another way to set eggs as being able to be put into the Incubator instead of being tag-based. I guess maybe incubation time could also be set with a recipe json. Main reasoning behind this is to make a recipe book for the incubator be possible.

Burning-Cactus:


Burning-Cactus:
Closing this for now since I'd like to restart on this.

PR #543, opened 2022-04-05
refactor: Zanite weapon math improvement.

### LINKED ISSUE #540: Enhancement: Adjust Zanite Weapon math again.
It should probably subtract whatever the base damage of the weapon is after the math is complete, instead of a constant 6 damage (which is the base damage of swords assuming no edits have been made).

bconlon1:
- [x] Closes #540.


Burning-Cactus (approved):
Looks good to me!

PR #546, opened 2022-04-08
fix: Use getMaxBuildHeight instead of getHeight for where to place the player when falling out of the Aether.

### LINKED ISSUE #541: Bug: Falling out of Aether spawns player at around y 378

System

### Other Type

_No response_

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

40.0.40

### The Aether 1 Version

1.0.0

_No response_

### Client Log

https://gist.github.com/ZolaKluke/be1f21f159ec5353e1bcdb9e1fc0ce23

_No response_

### Steps to Reproduce

1. Go into the Aether with a parachute
2. Fall out of the Aether
3. Look at y value when entering overworld

### What You Expect To Happen

Spawn at the dimension's set max height

### What Actually Happened

Spawned higher than what I should've

### Additional Details

Told to report by Bconlon

bconlon1:
- [x] Closes #541.


Burning-Cactus (approved):
Looks good to me!

PR #562, opened 2022-04-17
refactor: Capability syncing system.

### LINKED ISSUE #187: Bug: EntityDataManager does not like me storing data variables in the player capability.

bconlon1:
- [x] Closes #187.


Burning-Cactus (approved):
Looks good to me!

PR #571, opened 2022-04-18
feat: Implemented config entry validation for ConfigFilter codec

### LINKED ISSUE #568: Bug: Extra check to avoid crashes with usage of anything using ConfigSerializer.
A search for a config value from a path is not guaranteed to return a ConfigValue<Boolean>. We need an extra check for `ForgeConfigSpec.ConfigValue<Boolean> instanceof ForgeConfigSpec.BooleanValue` in the deserialize method.

Drullkus:
The type of value coming from the config ID should not be trusted. Java erases generics after compile, meaning the code executed during `ConfigSerializer.deserialize` will not error, even if the generic type is not Boolean. Otherwise, it will error and crash, while executing `this.config.get()` during Worldgen.


bconlon1 (approved):
lgtm

PR #572, opened 2022-04-18
Fix: Hopefully fixed the capability desync issues.

bconlon1:
Not marking this as closing anything until testing can be done on ANTS.


Burning-Cactus (approved):
Looks good to me!

PR #577, opened 2022-04-21
Update zh_cn translation

Parallel011:
Add more translation to match en_us, including Pro tips and the Book of Lore.


bconlon1 (approved):
lgtm.

PR #587, opened 2022-05-08
refactor: Improved DimensionType tag syncing for datapack usage.

bconlon1:
- Custom dimension tag syncing should now sync custom dimension tags not created by the Aether.

### DIFF
--- a/src/main/java/com/gildedgames/aether/core/util/LevelUtil.java
+++ b/src/main/java/com/gildedgames/aether/core/util/LevelUtil.java
@@ -1,6 +1,5 @@
 package com.gildedgames.aether.core.util;
 
-import com.gildedgames.aether.common.registry.AetherTags;
 import com.gildedgames.aether.core.AetherConfig;
 import com.gildedgames.aether.core.network.AetherPacketHandler;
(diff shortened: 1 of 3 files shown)


Drullkus (changes requested):
Largely looks good, just one small suboptimal detail to address

src/main/java/com/gildedgames/aether/core/util/LevelUtil.java
```diff
@@ -54,7 +52,7 @@ public static boolean inTag(Level level, TagKey<DimensionType> tagKey) {
         return Boolean.TRUE.equals(tagTrackers.get(dimensionKey, tagKey));
     }
 
-    public static List<TagKey<DimensionType>> getTags() {
-        return tags;
+    public static List<TagKey<DimensionType>> getTags(Level level) {
+        return level.dimensionTypeRegistration().getTagKeys().toList();
```
Drullkus: You should limit the tags sent to the client, to the Aether namespace to avoid potentially sending the client an excess of Dimension Type tags that are otherwise never used

Drullkus (approved):
Perfection!

PR #591, opened 2022-05-10
Fix: Mobs can no longer fall out of the Aether when not attached to a player.

### LINKED ISSUE #535: Bug: Aether mobs are falling into the Overworld.
- [ ] Cockatrices
- [ ] Moas
- [ ] Whirlwinds
- At this point it might just be worth it block off mobs that aren't mounted from falling to the Overworld. However, it's still a good idea to go over the AI of the above mobs.

Burning-Cactus:
- [x] Closes #535


bconlon1 (approved):
lgtm

PR #594, opened 2022-05-14
feat: Potted Bush Stems and Berry Bushes.

### LINKED ISSUE #74: Enhancement: Allow berry bushes to be placed in flower pots
This is the only non-full-cube (sorta) plant type that can't be placed in pots (all flowers and saplings can). Allowing them to be added to pots would give just that bit more capability for customization. May require a dedicated texture, though, as having the full stem render in the pot might look a bit clunky - I might make such a texture once aether textures are revamped for 1.14+.

Perhaps a bit out of place for vanilla, but quark allows sweet berry bushes to be placed in pots, so maybe this could be only allowed if quark is present. At the same time, I don't see much point in making the presence of other mods necessary for stuff like this, so it could be a default feature.

bconlon1:
- [x] Closes #74.


Burning-Cactus (approved):
Looks good to me!

PR #613, opened 2022-05-17
feat: additional texture reworks and animations

OzPayn:
fixed fast graphics leaves, added water animation to aether portal and portal item, adjusted contrast on ore textures


bconlon1 (approved):
lgtm

PR #615, opened 2022-05-17
Added Inebriation to "How did we get here" advancement

Mr-Aerod:
I just took the vanilla data/minecraft/advancements/nether/all_effects json file and added the inebriation effect in.
I hope this is the way it should be done. I know some datapack stuff, but very little to none about modding tbh.


bconlon1:
Not ideal; this would conflict with any other mod or pack that overrides the advancement too.

PR #623, opened 2022-05-19
Aether 1: French Translation

Aethyus:
[fr_fr.zip](https://github.com/Gilded-Games/The-Aether/files/8730300/fr_fr.zip)
.
Here is the French translation of Aether 1. Contrary to the one of 1.12.2, it contains all the new additions, but also translates the Book of Lore, which until then, was not translated into French, in the old versions of the mod.


bconlon1:
Implementing through #625.

PR #635, opened 2022-05-23
feat: model textures

OzPayn:
Reworked and updated all textures in the /textures/models folder (armor, capes, gloves, pendant, repulsion_shield)


bconlon1 (approved):
lgtm

PR #636, opened 2022-05-24
Feature: Valkyries

### LINKED ISSUE #207: Feature: Thunder Crystal Projectile
- Dependent on #223.

### LINKED ISSUE #222: Feature: Valkyries
- [x] Model, animation, and rendering.
- [x] Entity behavior, AI (teleportation), main entity class code.
- [x] Valkyrie teleportation can probably be based on finding valid angelic stone blocks so the Valkyrie doesn't teleport out of the Silver Dungeon.

### LINKED ISSUE #223: Feature: Valkyrie Queen
- [ ] Remove the Halo from the Valkyrie Queen model and texture.

Burning-Cactus:
- [x] Closes #222 
- [x] Closes #223 
- [x] Closes #207 
- [x] Implement new target goal for aggro


bconlon1 (changes requested):
- If you hit a valkyrie while its facing away from you, it will try to lunge at you while still facing away from you.

bconlon1 (changes requested):
- Valkyries shouldn't lose aggro on the player.
- Valkyries should be able to lunge multiple times, not just when the player first hits them and they're in the air. There should probably be some delay to this though so it's not instant like it was in 1.2.5 so players get time for weapon cooldown.
- Valkyries should be able to fly towards the player when in the air, and also jump during pursuit.
- Valkyries shouldn't despawn in peaceful; they should just be passive/unprovokeable.

bconlon1 (changes requested):
- [x] "I'll return when I have them." Dialogue line for getting more Victory Medals doubles up.
- [x] Valkyrie Queen should also drop a Golden Sword.
- [x] Valkyrie Queen should be immune to fire.
- [x] Valkyrie Queen can walk away while mid-dialogue.
- [x] Valkyrie Queen Boss Bar could be offset downwards slightly.
- [x] Split layer definitions for Ice Crystal and Thunder Crystal instead of having one Crystal definition.
- [x] Thunder Crystals can be hit back which I think was only a thing for Ice Crystals. 
- [x] Thunder Crystals should probably spawn lightning when they die from running out of timer.
- [x] Thunder Crystals might have weird behavior with going invisible during the fight, which seems to possibly correlate to them colliding with the Queen, but they also might still exist because the player can still be hit by lightning afterwards, as if the crystal were invisible.

bconlon1 (changes requested):
- [x] Thunder Crystals can collide with the Valkyrie Queen.
- [x] Thunder Crystals should be one hit to destroy.

bconlon1 (approved):
lgtm

PR #637, opened 2022-05-25
feat: Gravitite Jump now uses keybind.

### LINKED ISSUE #634: Enhancement: Gravitite Armor's high jump should likely be off by default.
Make it shift jump with a configurable keybind.

bconlon1:
- [x] Closes #634.


Burning-Cactus (changes requested):
This technically works on a dedicated server, but only because movement is handled on the client. The key binding can't be accessed from the server, so the game fails to find it and prints some errors in the log.

Burning-Cactus (approved):
Looks good to me!

PR #638, opened 2022-05-26
refactor: Improved Aerwhale model code.

### LINKED ISSUE #495: Chore: Redo entity model rotations.
Currently, the aerwhale model renders perfectly fine, but the cubes are oriented so that it looks like the aerwhale is facing right. It's easy enough to fix this by rotating the pose stack in the aerwhale's renderer, but with entity rendering changes possibly on the horizon, it may be better to address this at some point.

- [x] Aerwhale
- [x] Parachute
- [x] TNT Present

bconlon1:
- [x] Closes #495.


Burning-Cactus (dismissed):
The aerwhale needs to be looking in the direction it's moving in.

Burning-Cactus (approved):
Looks good to me!

PR #639, opened 2022-05-26
fix: Fixes to render rotation for Parachutes and Presents and also XP spawn position from Presents.

bconlon1:


Burning-Cactus (approved):
Looks good to me!

PR #647, opened 2022-05-31
feat: entity textures

OzPayn:
all of them I think


OzPayn:
The new Aerwhale and Sun Spirit models need to be implemented before this can be merged

OzPayn:
Oh actually its still using the old Aerwhale layout. I'll commit the new one real quick as I expect this to be merged after the new model is merged.

bconlon1 (approved):
lgtm

PR #650, opened 2022-06-01
fix: Various Aether actions will no longer kick you for flying.

### LINKED ISSUE #433: Bug: Investigate kicks for flying.
ServerGamePacketListenerImpl has the code for handling these kicks, it seems to be handled if clientIsFloating is true, which is determined around line 878. The only easily modifiable value that seems to influence this is player.verticalCollisionBelow, but this seems to be the best way considering its only usage is in these kick checks.

Might be able to set a tracker or something for if the player is doing certain mod actions, to not kick them for flying.

- [x] Parachutes
- [x] Valkyrie Armor
- [x] Slowfall Accessories
- [x] Flying Animals

bconlon1:
- [x] Closes #433.


Burning-Cactus (approved):
Looks good to me!

PR #655, opened 2022-06-04
Fix: Weather command works and weather resets when sleeping in the Aether.

### LINKED ISSUE #632: Bug: Issues with weather.
- [x] Setting rain from the Aether doesn't seem to work properly, but setting thunderstorms does.
- [ ] Thunderstorms darken the Aether too much (may not be fixable light level-wise)
- [ ] Skybox could maybe have improvements if possible to darkening colors during rain.

### LINKED ISSUE #633: Bug: Sleeping during Thunderstorm
You can sleep in the Aether during a thunderstorm (indicated by the darkened skies) but it does not clear the weather like it should. Could be related to other weather related issues.

Burning-Cactus:
This pull request addresses some of the quirks with how the Aether handles its weather cycle.
- [x] Closes #632 
- [x] Closes #633


bconlon1 (approved):
lgtm

PR #657, opened 2022-06-04
feat: Fog distance tweaks.

### LINKED ISSUE #333: Bug: Strange Cloud Fog behavior
image

When quickly changing your player's moment delta, it seems like the fog will arbitrarily jump around on the top surface of the clouds.

It's probably something with the renderer. I'll get to the bottom of it

### LINKED ISSUE #456: Feature Request: Beta fog shader
Fog renderer that recreates how the Aether's fog looked in the original mod. Not sure how doable this is.
- Fog at a distance.
- Fog during the rain.

I think the main thing that made the beta fog was it had more distance to it, it faded out more while modern fog is more abrupt.

We might be able to also use the Forge fog render event to change the distance of fog without having to construct anything from the ground up.

bconlon1:


Burning-Cactus (approved):
Looks good to me!

PR #661, opened 2022-06-07
fix: Menu fixes.

### LINKED ISSUE #658: Bug: Menu bugs.
- [x] Fix menu rotation to be slower and also possibly based somewhat on rotating around the player instead of from the player's camera (this should be doable with setting F5 perspective and cancelling the player render event when the menu is active).
- [x] Fix splash resetting when entering any other screen from the menu then returning to the menu.

bconlon1:
- [x] Closes #658.


Burning-Cactus (approved):
Looks good to me!

PR #668, opened 2022-06-10
feat: Moa and Moa Egg fixes as well as handling for Moa healing.

### LINKED ISSUE #667: Bug: Moa type briefly flashes incorrectly when using the colored moa eggs as a spawn egg.
Visibly looks as if the moa type is being set after its spawned.

bconlon1:
- [x] Closes #667.


Burning-Cactus (approved):
Looks good to me!

PR #670, opened 2022-06-12
fix: Worldgen fixes.

### LINKED ISSUE #271: Enhancement: Switch any hardcoded block values used by features to determine valid generation positions to be tags instead.

### LINKED ISSUE #273: Enhancement: Configure Golden Oak trees so that they always spawn with branches.
May require a custom tree generator since vanilla's fancy oak generator seems to be capable of spawning without branches.

### LINKED ISSUE #519: Bug: Skyroot trees generating on Crystal tree islands

_No response_

### Other Type

_No response_

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

 40.0.19

### The Aether 1 Version

1.0.0

_No response_

### Client Log

_No response_

_No response_

### Steps to Reproduce

1. Generate a new world
2. Load it and try to find one.

### What You Expect To Happen

I don't expect Skyroot trees being able to generate on Crystal tree islands.

### What Actually Happened

Guess what

### Additional Details

_No response_

### LINKED ISSUE #620: Bug: Issues with lakes.
- Lakes in the Aether generate a barrier of grass blocks around them even where there should be dirt or holystone.
- It seems to be possible for lakes to either cut off trees or have trees generate in them. The former is probably fine since I believe that's a vanilla bug however the latter probably shouldn't be able to happen but that may be a bug with trees.

bconlon1:
- [x] Closes #273.
- [x] Closes #519.
- [x] Closes #620.
- [x] Closes #271.


Burning-Cactus (approved):
Looks good to me!

PR #680, opened 2022-06-18
Update zh_cn translate.

Parallel011:
Adjust some new translation.
Also added more translate to match en_us.


bconlon1:
Remake this PR to the 1.19 branch.

PR #732, opened 2022-07-21
refactor: Entity Fixes + Deferred API Registries

### LINKED ISSUE #731: Bug: Error log when trying to /summon Moa.

bconlon1:
- [x] Closes #731.
- [x] Parachute behavior more accurate to original mod (e.g. rotation and speed, sound and other behavior).
- [x] Switch API registries to DeferredRegister.

### DIFF
--- a/src/main/java/com/gildedgames/aether/entity/miscellaneous/Parachute.java
+++ b/src/main/java/com/gildedgames/aether/entity/miscellaneous/Parachute.java
@@ -6,7 +6,6 @@ import com.gildedgames.aether.util.EntityUtil;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.network.protocol.Packet;
 import net.minecraft.server.level.ServerPlayer;
-import net.minecraft.util.Mth;
 import net.minecraft.world.entity.LivingEntity;
 import net.minecraft.world.entity.ai.attributes.AttributeInstance;
 import net.minecraft.world.phys.Vec3;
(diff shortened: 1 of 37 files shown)


src/main/java/com/gildedgames/aether/entity/miscellaneous/Parachute.java
```diff
@@ -53,45 +50,60 @@ public void tick() {
...
+        float speed = 0.03F;
+        this.moveRelative(speed, vec3);
+        this.move(MoverType.SELF, this.getDeltaMovement());
+        return this.getDeltaMovement();
+    }
+
     public void spawnExplosionParticle() {
-        for (int i = 0; i < 1; ++i) {
+        if (!this.level.isClientSide) {
+            AetherPacketHandler.sendToAll(new ExplosionParticlePacket(this.getId()));
```
Burning-Cactus: This probably doesn't need to be sent to all clients.

src/main/java/com/gildedgames/aether/entity/miscellaneous/Parachute.java
```diff
@@ -53,45 +50,60 @@ public void tick() {
...
+        } else {
             EntityUtil.spawnMovementExplosionParticles(this);
         }
     }
 
     public void die() {
+        for (int i = 0; i < 10; i++) {
```
Burning-Cactus: We should probably stop sending packets in loops, and instead include a number of iterations that the client should go through in a loop so we minimize the number of packets sent.

Drullkus (dismissed):
I don't understand what `DungeonType`s are for. These are redundant since these can already be 1:1'd to the structure system

Burning-Cactus (changes requested):

Burning-Cactus (approved):
Looks good to me!

PR #736, opened 2022-07-31
feat: Data-driven Incubation.

### LINKED ISSUE #529: Enhancement: Convert Incubator behavior to be recipe-based.

bconlon1:
- [x] Closes #529.


Burning-Cactus (approved):
Looks good to me!

PR #744, opened 2022-08-13
fix: Icestone performance improvements.

bconlon1:


Drullkus (approved):
That's a cool solution, well optimized!

PR #745, opened 2022-08-14
feat: 1.19.2 Port

bconlon1:
- Update to 1.19.2.
- Change mod version to 0.0.1 instead of v1.0.0.
- Add new tags based on recent forge tag additions.


Drullkus (approved):
LGTM!

PR #766, opened 2022-08-19
Silver Dungeon NBT files

reetamb:
These will need processors to change chests into mimics, and Angelic Stone blocks into their Light and Trapped variants.
We will also need a way to attach Aercloud features onto the bottom of them.

Every single structure component shares a wall. The boss_room shares its walls with the boss_exterior, which shares its front wall with the back wall of the entrance_exterior, which shares its interior walls with the empty_room and chest_room templates, who all share walls with themselves and each other. Whew, that was a mouthful.

Door openings don't exist on the interior, because they will need to be generated in code somehow to make them more randomized.


bconlon1 (approved):
lgtm

PR #779, opened 2022-08-27
refactor: Fixes and Improvements

### LINKED ISSUE #767: Issues with Tips
Hello, I'm here to point out a proposal/problem that would correct many things in the French translation of the mod. Currently, the "Pro Tips" that are displayed are always on one line, which is a bit annoying. The problem is that I can't really translate the original sentences without them overflowing/not fitting in the screen, which happens 80% of the time. So I propose a system where every time a sentence is longer or shorter, it is put on several lines, to avoid this kind of bugs.

### LINKED ISSUE #771: Bug: Bypass block functionality for our blockstate recipes isn't working.

### LINKED ISSUE #775: Enhancement: Datagen classes.
Need to see if AetherDataGenerators can be simplified to be static and use method generics to avoid having to create a new object for each data generator.

bconlon1:
- [x] Closes #775.
- [x] Closes #767.
- [x] Closes #771.


Burning-Cactus (approved):
Looks good to me!

PR #784, opened 2022-08-29
Locked Silver Dungeon

reetamb:
Silver Dungeon blocks are Locked as necessary. Also fixed an oversight where there was some cobble left in one of the NBT files (oops)


bconlon1 (approved):
lgtm

PR #786, opened 2022-08-30
Feat: Invisible Blocks

### LINKED ISSUE #113: Bug: Dungeon structures need protection blocks properly reimplemented.
In twilight forest mod, some structures are protected - removing blocks results in them being placed back and some green sparkels

I think it makes sense to apply this to aether aswell, so only blocks that are meant to be broken in dungeons can be broken.

Burning-Cactus:
- [x] Closes #113


bconlon1 (approved):
lgtm

PR #789, opened 2022-08-30
Sun spirit fixes

Burning-Cactus:
- The sun spirit no longer takes hits from his own ice crystals.
- Fire minions are now immune to the sun spirit's attacks.
- Reverted the sun spirit to his original width.


bconlon1 (approved):
lgtm

PR #795, opened 2022-09-01
Slider Fixes and Improvements

### LINKED ISSUE #787: Bug: A few bugs with the Slider Boss Fight

Entity

### Other Type

_No response_

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

43.0.19

### The Aether 1 Version

1032436

_No response_

### Client Log

_No response_

_No response_

### Steps to Reproduce

**ISSUE 1**  - The Slider will sometimes not move, but still play the sliding sound:
I'm not _exactly_ sure how this happens, but I have noticed it happening the most when the slider is on the opposite side of the room from the player and is less than 1 block above the ground. Jumping near the Slider (next issue) can be used to sometimes get it under 1 block above the ground.

-----

**ISSUE 2** - Jumping near the slider causes it to think the player is above it (which they _technically_ are, but only the position) and to move upwards (can cause issue 5, and possibly issue 1 as well):
1. Spawn and begin fighting the Slider
2. Jump nearby it while avoiding its attacks

I have an idea for how to fix this. It is by making the slider move upwards if the bottom of the player's hitbox is above the top of the slider's hitbox, instead of based on the position. To (hopefully) also fix issue 5, this could be only done to the _upward_ movement, and not the others (or at least not to the downward movement)

-----

**ISSUE 3** - Running and Jumping around the slider confuses it and makes it not move much, which makes the fight way easier (related to issue 2):
1. Spawn the slider and begin the boss fight
4. Jump around the slider in circles, while staying somewhat close to the slider
---
**ISSUE 4** - Slider resets when the player touches the edge of the room:
1. Spawn the Slider
5. Hit it with a pickaxe
6. Run to and touch the wall of the room (it doesn't do it with every wall)
---
**ISSUE 5** - Slider doesn't move back to the floor if it is slightly above the ground (this one isn't as much of an Unexpected Behavior, but more of an Annoyance that has annoyed me for a long time, in several different versions of the Aether and Aether II):
1. Jump around an awakened Slider (in suvival mode)
4. It will eventually move upwards, and stay floating

### What You Expect To Happen

Issue 1: I expected for the Slider to move without getting stuck in the air

Issue 2: I expected for the Slider to stay on the ground since I wasn't high enough for it to not be able to hit me

Issue 3: I expected for the Slider to not just stay mostly still, and (an idea for a fix I had) for it to move farther, which could either hit me, or break out of the spot it was cornered at

Issue 4: I expected for the Slider to not reset the bossfight

Issue 5: I expected for the Slider to move back to the ground when hovering only slightly above the ground

### What Actually Happened

Issue 1: The slider got stuck in mid air, and played the slide sound without moving

Issue 2: The slider moved upwards when I jumped near it

Issue 3: The slider couldn't move much

Issue 4: The boss fight reset when I touched the wall

Issue 5: The slider hovered above the ground, and didn't move back down

### Additional Details

**ISSUES 1 & 5:**

https://user-images.githubusercontent.com/60141811/187327909-7e44559b-68c5-4460-b516-59261bbce954.mp4

---

**ISSUE 2:**

https://user-images.githubusercontent.com/60141811/187328328-240d9e3e-8abc-4e17-8775-9a0a5a3dfed1.mp4

---
**ISSUE 3:**

https://user-images.githubusercontent.com/60141811/187328750-cf949e9f-d507-4cbd-a744-30a8f1f4295f.mp4

---
**ISSUE 4:**

https://user-images.githubusercontent.com/60141811/187328855-69ffcb8b-307c-458f-9982-a4cecaa2d862.mp4

Zepalesque:
- [ ] Fixes #787
-
---
Changes:
- The Slider stays on the dungeon floor if the player is not high enough for it to need to move (fixes issues 2 and 5 from the bug report)
- The Slider will align with four blocks on the floor when spawned with a spawn egg, because previously it spawned on the center of the block, which caused the actual dungeon to be offset from where the blocks were. This fix will also make it easier to respawn the slider in the generated bronze dungeons, once they are added  (fixes issue 4 from the bug report)
- The Slider AI code is easier to read/understand (separated into multiple methods with explanations of what they do), and is simpler in some spots
- The non-pickaxe tool message timer cooldown works correctly (before the timer only decreased when hit, and the message only played again after being hit 15 times)
- The Slider no longer tries to break the dungeon floor, which caused it to get stuck in the air (fixes issue 1 from the bug report)
- The Slider no longer tries to push creative players upwards, which resulted in bubble-column-like movement
- There are a few things to make it easier for addons to make new Slider AI without using lots of reflection
- Issue 3 from the bug report is not fixed, because I couldn't figure out how to make it work correctly, and I realized that it _probably_ doesn't _need_ to be fixed, and it's not really any more of an exploit than aoiding its attacks the conventional way of circing around it slower
---
Fix demonstration (**The resource changes are _not_ part of the pull request, I was using a resource pack while recording these**):

- Jumping around the Slider no longer makes it move upwards:
https://user-images.githubusercontent.com/60141811/187801018-355bae72-d0a7-4a73-8890-e9a6bc2d1a63.mp4

- The slider moves to the floor instead of the player's foot level if they are low enough for the slider to reach
https://user-images.githubusercontent.com/60141811/187801298-d7a82b9d-6e11-4286-a1f7-3553f7175e73.mp4

- The slider tries to stay on the floor:
https://user-images.githubusercontent.com/60141811/187801426-47940dd8-f59a-4f53-8dda-130465328a6a.mp4

- Slider spawn position alignment (Pay attention to where the block outline is):
https://user-images.githubusercontent.com/60141811/187788900-21d736af-78ab-4199-a302-a231a9b117b4.mp4

- Message cooldown **(the message isn't changed, that's from the resource pack)**
https://user-images.githubusercontent.com/60141811/187794025-32d75f77-2d6a-4e23-8983-fa443dd86516.mp4

---
I agree to the Contributor License Agreement (CLA)


bconlon1 (changes requested):
Likely due to the changes to fix issue 1 of #787, the Slider now gets confused trying to navigate to the player if they are standing one block higher than the floor that the Slider is on; the Slider tries to move towards the player but can't move upwards because the player isn't above the Slider's hitbox. There also seems to be an issue with the Slider not being able to navigate to the player if the player is diagonal from the Slider and one of its directional paths is blocked, however this issue may have been an oversight on my end when originally writing the code.

It's important to remember the final boss room will have the platform for the treasure room entrance in the middle meaning there will be natural height variation that will have to be accounted for.

Zepalesque:
In the latest commit, I somehow accidentially _deleted the slider_ without realizing until a while after I made the commit, and I won’t be able to fix it for a few days, but the previous commit, [e66973e](https://github.com/Gilded-Games/The-Aether/pull/795/commits/e66973e1d7736d57622f347e002dc563a5732eda), has the problems with the platform and one-block upward movement fixed.
There are still a few more things that I want to fix, including redoing the modified slider spawn egg behavior I had added and then removed in the latest commit, but the movement is pretty much done in the latest commit, but if there are any more issues I will be able to fix them in a couple of days.

Here is the fix for the center platform issue:

https://user-images.githubusercontent.com/60141811/187997778-5acc7001-8f0e-4b87-8936-8cd6bcabd372.mp4

With this fix, I think it might also be a good idea to make the slider slightly faster, since it sometimes has to move up to go over the pedestal, which can make it take longer to reach the player, but that probably would be fixed when the boss balancing happens, so I didn’t change it since it will be rebalanced later anyway.

bconlon1 (changes requested):
I think also if you're going to make it so the Slider's fields are modifiable for addon compat, keep them private but add getters and setters.

Zepalesque:
I reverted the commit which deleted the slider file, fixed the rest of the things I wanted to fix, and made those fields private, so as long as there aren't any more issues or bugs, this is done now.

bconlon1 (changes requested):
- I don't particularly like the solution of fixing detection for moving over obstacles by hardcoding to detect that there is a central platform, if a custom dungeon is set up without a central platform but still has other obstacles of differing height then the Slider can't navigate them.
- There seems to be an issue where for some reason the Slider is hard to activate when attacked, like it just resets immediately more often than not, and also it seems to reset in creative mode which is not how it should work.
- I'm not sure what the point is of the custom Slider spawn egg is since it seems like the position setting could just be done in the Entity similar to how you've partially already done.

Zepalesque:
For the spawn egg, the custom item is so that the slider will spawn on the intersection of blocks closest to where the player clicked, and the mob doesn't know where the player clicked to spawn it, so it can't be set in the entity class.
Like this:
https://user-images.githubusercontent.com/60141811/188497658-904d4b78-1c28-4c21-94e0-098e15dffe53.mp4
so that these (when clicked on the corner closest to the middle of the platform) will all spawn on the center of the platform, similar to how doors are placed with the hinge on the side of the block the player clicked on:
https://user-images.githubusercontent.com/60141811/188498283-89efac5e-43ff-4ce1-88fa-b0e67df5c68f.mp4

With the hardcoded spawn platform issue, would adding an NBT tag which can have any number of bounding boxes which the slider tries to go over work as a fix? (I haven't done this yet, since I don't know if that's what you mean or not, but if it is then I will do that soon)

bconlon1:
I see, the custom spawn-egg makes sense then. For the spawn platform thing I'd personally just not use any checks for bounding boxes at all and instead have the slider be able to scale any obstacles if it can if it just detects it is hitting a block or some other checks. The only potential solution I can think of though to making that work while also not having the Slider move up in any instance where the player is above it (like jumping) is to have it be timer based, like if the player's minY is above the middle of the Slider's hitbox for a certain amount of time, then it will choose to move upwards, and if its below it could automatically default to moving downwards. However I guess there is still the issue in that of the Slider being able to choose whether to move horizontally or vertically so it doesn't get stuck going down into the floor if the player is at a lower elevation. Could be collision based to run the timer or something.

Zepalesque:
There are a couple of problems with doing it that way that I'm not sure how to fix (this is starting to go past my current modding/coding knowledge, so trying to figure out how to code some parts of this is getting pretty difficult):
- If the player goes around the center platform (or any obstacle) instead of over it, and stays on the ground the whole time, the slider wouldn't know that it should go above that obstacle, since the player never went above halfway to the top of it's hitbox.
- I'm not sure how to make the slider know when to stop going up, for obstacles taller than one block.
- If a custom dungeon has pillars then the slider wouldn't know how to go around them.
- If it goes up based on collision, then it would be slower and take more movements to go over something, because it would do something like this: _|¯¯| (4 movements) instead of this: |¯¯¯| (3 movements)

bconlon1:
Maybe it'd be better then to get rid of the platform functionality for this PR, we can look into improving any other parts of the Slider AI that have been discussed but not fixed in this PR at a later date.

Zepalesque:
Ok, I removed it for now. Also, I found something that _might_ work with the slider for navigation (but I'm not 100% sure): PathNavigation and PathFinder, since the slider already extends PathfinderMob (I haven't looked a _lot_ at them, but I think they could possibly be useful)

bconlon1:
We attempted to look into path navigation previously but it was hard to get working with the Slider's unique movement so we went with using the old method of AI they used.

Zepalesque:
I changed everything else I still wanted to, so as long as there aren't any bugs I missed (which I didn't find any of while playtesting, so there _hopefully_ aren't any more of), this is complete now.

bconlon1:
I'll try getting around to re-reviewing this during the weekend.

bconlon1 (approved):
lgtm

PR #808, opened 2022-09-01
refactor: Use suppliers to keep track of moa types

### LINKED ISSUE #806: Enhancement: Please use broader Object types for external compatibility
What's up gamers, based Modding Legacy associate and Cryptic Mushroom ring leader here.

Anyways, here's the entire issue:

Java gaming

Obviously this is more of a dev style nitpick than anything else, but it will be helpful for outside creators when creating addons for the mod. It's kind of an unfortunate side effect of maintaining a very popular mod that is also open-source. Another solution would be to expose a public API, but it's too much of a hassle since the entire mod is open source anyway.

bonus points if you use "ALT + ENTER" on a method signature's return or argument type, there's a shortcut that recommends a bunch of types that extend the existing type (for example if you alt + enter on `RegistryObject<?>` in a `public static RegistryObject<?> test()` method, there's an option to change the return type and IntelliJ will then recommend `Supplier<?>` and `Object`.

ok back to music school bye bitch

Jonathing:
Fixes #806.

Uses type of `Supplier<? extends MoaType>` rather than `RegistryObject<MoaType>` for keeping track of moas. This has two benefits:

1. Other mods can use their own suppliers to make their own moa type eggs.
2. Suppliers or RegistryObjects can be of a subtype of MoaType without needing to resort to unnecessary casting.

PR editing has been enabled in case you want to build off of what I provided.

### DIFF
--- a/src/main/java/com/gildedgames/aether/entity/monster/Cockatrice.java
+++ b/src/main/java/com/gildedgames/aether/entity/monster/Cockatrice.java
@@ -212,11 +212,6 @@ public class Cockatrice extends Monster implements RangedAttackMob, WingedBird, 
     }
 
     @Override
-    protected float getSoundVolume() {
-        return 0.25F;
-    }
-
-    @Override
     protected int calculateFallDamage(float distance, float damageMultiplier) {
         return 0;
     }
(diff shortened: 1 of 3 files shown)


bconlon1 (approved):
lgtm

PR #809, opened 2022-09-01
refactor: Defer to using constructors directly rather than Google's helper methods

Jonathing:
The usage of the classes `Lists`, `Maps`, and `Sets` is a relic of the past when Java's generic type system was confusing and unfinished. The use of methods like `Lists.newArrayList()` are now no longer necessary (as stated by Google themselves in the developer comments) and the direct constructors of these objects should be used. This will help keep unnecessary calls off the stack and can keep compatibility with forward Java versions. Additionally, these methods have proven somewhat problematic on the Linux operating system in certain scenarios.


bconlon1 (dismissed):
lgtm

Jonathing:
how

bconlon1 (approved):
lgtm

PR #812, opened 2022-09-02
Moa Volume is no longer too low

### LINKED ISSUE #803: Bug: Moa volume is too low.
Needs to be increased in the Moa entity code.

reetamb:
Fix #803 

The Moa and Cockatrice no longer override volume, so they'll have the default volume of Living Entities (1.0F).


bconlon1 (approved):
lgtm

PR #823, opened 2022-09-06
feat(compat): JEI Compatibility

### LINKED ISSUE #289: Feature: JEI compatibility
- [x] Altar and Freezer recipes, with support for displaying custom fuels.
- [x] Incubators should have support to display eggs that can be incubated.

saintselect:
```
Adds JEI compatibility for the Altar, Freezer and Incubator.

Closes #289
```


saintselect:
Showcase:

bconlon1:
Technically this PR does not close #289 considering the potential of us adding a recipe category for dungeon loot, but I still need to discuss that with other team members (if we decide against it I can just close the issue in the future after this is merged). Otherwise though this looks good to me! (Will approve later in the day once I'm back at my PC).

bconlon1:
(Actually we're probably gonna leave the dungeon loot thing to JER so yeah this closes #289)

bconlon1 (changes requested):
Actually looking at this further compared with JEI, I've noticed a couple things:
- The GUI frame used in this PR should be the same as the one used by JEI, note that it has softer outlines and a smaller size. It also spaces itself out on the GUI differently.
image

- The frame should display time the recipe takes as well as any XP output
image

- There should be support for displaying what fuels the Aether's crafting stations use.
image

saintselect:
Oops closed it by accident, I'll reopen with the requested changes shortly.

saintselect:
Updated with the requested changes. Looks a lot nicer so thanks for the feedback. :)



---
I added all of the fuel types to a single category to make things more organized. It pulls valid fuel items from the associated block entities, so the JEI plugin should work just fine if the list of valid fuels is changed by an addon, etc. Also scales the menu based on the length of the translatable components, so it should look fine on other languages.

bconlon1 (approved):
lgtm!

PR #829, opened 2022-09-08
Recipe ID based grouping for vanilla blocks

### LINKED ISSUE #462: Enhancement: Find a better way of adding groups to vanilla recipes that doesn't involve replacing.
We replace some Minecraft recipes to add a group to them so they can be grouped with skyroot-crafted equivalents in the recipe book. This is probably a compatibility issue waiting to happen.

Might be able to do this if we mess around with the recipe book code a bit, not sure.

We could also just give up on this and not bother.

Burning-Cactus:
Closes #462


bconlon1 (approved):
lgtm

PR #837, opened 2022-09-10
fix: Accessory screen updates its width value dynamically according to gamemode

### LINKED ISSUE #820: Bug: Accessory Inventory doesn't properly update its horizontal size.
When entering creative and having the extra space for the delete button display, it does not properly shift the edge of the GUI in code. We have a workaround for this with effects but it turns out this affects other modded GUIs like JEI's, so the effect solution will have to be removed as its only a patchwork one.

bconlon1:
- [x] Closes #820.


Burning-Cactus (approved):
Looks good to me!

Issue #776, opened 2022-08-27
Enhancement: Boss balancing and fixes


bconlon1:
Slider: 
- [x] Slider crashes at https://github.com/Gilded-Games/The-Aether/blob/50c7d479aa5eb6a8ebda8d243d43301a6d4e110f/src/main/java/com/gildedgames/aether/entity/monster/dungeon/boss/Slider.java#L606
- [x] Slider can stun lock the player into a corner.
- [x] In normal mode, adjust delay time or speed very slightly to make the slider faster.
- [x] Maybe make slider randomly adjust its vertical position slightly to make it less predictable.
- [x] Make it not deactivate if out of range if it has no dungeon; however it shouldn't break blocks.
- [x] Weird edge case where slider will move up and down if it corners player into a wall but not enough to reach it to damage it.
---
Valkyries
- [x] Valkyrie lunge attack usually launches Valkyries behind the player and not at the player, allowing them to dodge shields by attacking from behind.
- [x] Valkyrie lunge needs more cooldown/Valkyries need some kind of pause during combat so they don't relentlessly attack.
- [x] Valkyries probably need better pathfinding when flying, alongside the Valkyrie Queen. They can fly away from the player to try to land in a far-off direction.
---
Valkyrie Queen: 
- [x] Thunder Crystal needs more consistent collision behavior.
- [x] Thunder Crystals should spawn lightning on death.
- [x] Needs lunge attack back. 
- [x] Verify that it can teleport up to players and get unstuck properly if cheesed.
- [x] Potentially change InBossRoomTargetGoal to be nearest player and not random player. If this is changed then Fire Minions should also use it.
- [x] Prevent Valkyrie Queen from riding boats.
- [x] Prevent Valkyrie Queen from teleporting outside the boss room; valid teleport position should check for that.
- [x] Valkyrie Queen can navigate outside of boss room.
- [x] Valkyrie Queen can be damaged and killed by her own lightning.

OzPayn:
- [x] - Sun Spirit needs separate hitboxes for ice crystal hit registration and player collision. [Visual example](https://i.imgur.com/4qgXpWK.png)
- [x] - Ice Crystal and Fire Crystal hitboxes need adjusting in the player's favour. [Example](https://i.imgur.com/gFRwOlz.png)
- [x] - Fire Minion hitbox needs to be wider for more reliable melee hits. 
- [x] - Fire Minion run speed needs lowering a little bit to account for slower overall attack output of the player
[...]

PR #840, opened 2022-09-11
Rebalance and fix the sun spirit

Burning-Cactus:
See #776


bconlon1 (changes requested):
- Fire Minions could be slightly faster.
- Fireball count is probably fine but fireballs could maybe fly a bit faster; try seeing whether or not its too overwhelming or not to make them spawn slightly more frequently though. Maybe they could have shorter lifespans to balance this out slightly though. But at that point it might not matter to make more spawn.
- Sun Spirit should place more fire at slightly more frequent random intervals.

bconlon1 (changes requested):
Also Fire Minions should have the same light overrides in their renderer as the Sun Spirit.

bconlon1 (changes requested):
Need to fix the Fire Minions not turning red when hurt.

bconlon1 (approved):
lgtm, just need Oz's review now.

PR #851, opened 2022-10-27
improv(recipe): Added blockstate recipe mcfunction support

00-Steven:
Adds mcfunction support to the existing BlockState recipe system.

---


src/main/java/com/gildedgames/aether/data/providers/AetherRecipeProvider.java
```diff
@@ -416,6 +417,10 @@ public BlockStateRecipeBuilder ambrosiumEnchanting(Block result, Block ingredien
         return BlockStateRecipeBuilder.recipe(BlockStateIngredient.of(ingredient), result, AetherRecipeSerializers.AMBROSIUM_ENCHANTING.get());
     }
 
+    public BlockStateRecipeBuilder ambrosiumEnchanting(Block result, Block ingredient, ResourceLocation mcfunction) {
```
bconlon1: This is probably not necessary considering it'd be unused by our data generators.

src/main/java/com/gildedgames/aether/recipe/builder/BiomeParameterRecipeBuilder.java
```diff
@@ -24,38 +24,62 @@ public class BiomeParameterRecipeBuilder extends BlockStateRecipeBuilder {
...
     }
 
     public static BiomeParameterRecipeBuilder recipe(BlockStateIngredient ingredient, Block resultBlock, Map<Property<?>, Comparable<?>> resultProperties, TagKey<Biome> biomeTag, BlockStateRecipeSerializer<?> serializer) {
-        return recipe(BlockPropertyPair.of(resultBlock, resultProperties), ingredient, null, biomeTag, serializer);
+        return recipe(BlockPropertyPair.of(resultBlock, resultProperties), ingredient, null, biomeTag, serializer, null);
     }
 
-    public static BiomeParameterRecipeBuilder recipe(BlockPropertyPair result, BlockStateIngredient ingredient, @Nullable ResourceKey<Biome> biomeKey, @Nullable TagKey<Biome> biomeTag, BlockStateRecipeSerializer<?> serializer) {
-        return new BiomeParameterRecipeBuilder(result, ingredient, biomeKey, biomeTag, serializer);
+    public static BiomeParameterRecipeBuilder recipe(BlockStateIngredient ingredient, Block result, ResourceKey<Biome> biomeKey, BlockStateRecipeSerializer<?> serializer, ResourceLocation mcfunction) {
```
bconlon1: The amount of methods could probably be cut down by having a new method to handle appending an mcfunction through the builder, similar to how vanilla does for things like groups:
```
public ShapedRecipeBuilder mcFunction(@Nullable ResourceLocation mcFunction) {
  this.mcFunction = mcFunction;
  return this;
}
```

bconlon1: Or well, doing that could probably be done in BlockStateRecipeBuilder instead which would cut down on even more changes.

src/main/java/com/gildedgames/aether/recipe/serializer/BiomeParameterRecipeSerializer.java
```diff
@@ -40,7 +41,7 @@ public T fromNetwork(@Nonnull ResourceLocation recipeId, @Nonnull FriendlyByteBu
         TagKey<Biome> biomeTag = BlockStateRecipeUtil.readBiomeTag(buf);
         BlockStateIngredient ingredient = BlockStateIngredient.fromNetwork(buf);
         BlockPropertyPair result = BlockStateRecipeUtil.readPair(buf);
-        return this.factory.create(recipeId, biomeKey, biomeTag, ingredient, result);
+        return this.factory.create(recipeId, biomeKey, biomeTag, ingredient, result, CommandFunction.CacheableFunction.NONE);
```
bconlon1: I believe you need to handle the networking for the function instead of it being empty or it may not work on multiplayer, but I'd need a datapack to test with to be certain.

bconlon1 (changes requested):
Most of this PR looks good aside from the attached requested changes. I have nitpicks like the capitalization of `mcfunction` or `Mcfunction` (but honestly I'm not sure what I'd do with it), and the usage of `var`, and it might be possible to abstract some of the code segments so they can be used through method calls instead, but I might just handle that myself since those aren't really implementation issues.

bconlon1 (approved):
lgtm

PR #854, opened 2022-11-07
improv(entity): Valkyrie and Valkyrie Queen improvements

Burning-Cactus:
```
Valkyries now have their lunge attack back.
Valkyries no longer get stuck in boats.
Valkyrie queens will teleport up to the player if the player is too high for them to reach for too long.

Partially addressed #776 
```


bconlon1 (approved):
lgtm

PR #855, opened 2022-11-08
fix(entity): Rework valkyrie lunge and increase slider dungeon bounds

Burning-Cactus:
```
Valkyries now take knockback into account instead of juggernauting through.
Bronze dungeons properly set the slider's dungeon bounds now.
```


bconlon1 (approved):
lgtm

PR #856, opened 2022-11-08
feat(block)!: Additions, fixes, and general cleanup of block code

bconlon1:
Cleaned up block package, blockentity package, portal code, Aether, AetherConfig, AetherTags, access transformers by switching to mixin accessors, and various mixins

BREAKING CHANGE: Various classes have been renamed and moved around


Burning-Cactus (approved):
Looks good to me!

PR #859, opened 2022-11-23
refactor: Balanced icestone fuel values

bconlon1:
Sets the Icestone Stairs, Slab, and Wall to the right fuel values in the Freezer


Burning-Cactus (approved):
lgtm

PR #864, opened 2022-11-27
fix(entity): Fix named easter eggs

### LINKED ISSUE #280: Feature: Easter Eggs
- [x] Verify all easter eggs are implemented.
- [x] Document some easter eggs as for some it is unclear what users they're actually meant to belong to, like Aerwhale riding.
- [x] Restore swuff_ cape as a cape item rename to try to continue maintaining easter eggs while staying within mojang's eula.
- [x] Make sure all the name-related easter eggs for entities work; Phygs don't work correctly and Fire Minions supposedly also do not. Moas probably work fine and if they do their method should be the standard.

bconlon1:
Fixes conditional checks for names to check between strings instead of checking a component with a string

closes #280


Burning-Cactus (approved):
Looks good to me!

PR #865, opened 2022-11-27
improv(block): Fix locked overlay applying to dungeon doorways

bconlon1:
The locked_dungeon_blocks tag is now only used for locked blocks, and the Slider now has its own separate tag for "locked" blocks it can't break which is slider_unbreakable


Burning-Cactus (approved):
Looks good to me!

PR #867, opened 2022-11-27
refactor: Clean up deprecation warnings

### LINKED ISSUE #299: Chore: Deprecation warnings.
See how many of them relate to update changes and are tied to newer methods that should be switched to.

EDIT: A lot of them seem to be Mojang's weird choice of using `@Deprecated` for certain methods. According to Forge this seems to have to do with `@Deprecated` annotations meaning only to not call a method, but not to avoid overriding it, so it doesn't mean that methods annotated this way should be avoided completely. I believe its related to stack vs item sensitive methods as well as state vs block sensitive methods.

bconlon1:
closes #299


Burning-Cactus (approved):
Looks good to me!

PR #877, opened 2022-12-03
fix: fixed silver heart rendering in certain scenarios

### LINKED ISSUE #822: Bug: Modifying generic.max_health attribute to the player breaks life shard silver heart overlay.

Gui

### Other Type

_No response_

### What Type of Bug Is This?

Annoyance

### Forge Version

43.1.1

### The Aether 1 Version

1.0.0

_No response_

### Client Log

_No response_

_No response_

### Steps to Reproduce

- Modify the player generic.max_health attribute in any way, either through mods like origins or using vanilla features like applying attributes to armor and then wearing that armor, or even the health boost potion effect
- consume life shards

### What You Expect To Happen

I expected the silver hearts to show up like how golden absorbtion hearts show up, dynamically adjusting for your actual red hearts count. You can notice what i mean by giving yourself absorbtion and then use health boost to increase this attribute.

### What Actually Happened

Instead, they are static, they dont change their location and apparently they are a clone of the newly added hearts, mirroring their damage status. 

### Additional Details

The issue can he noticed very easily when you mod your max health to values greater than 40, making you have three or more rows of red hearts or lower than 20, substracting from the first row. Im coming later with photos to better show this, but for now i have to do some homework :copium:

GizmoTheMoonPig:
Fixes silver heart rendering in a few scenarios.
- if the player has less than 10 base hearts, silver hearts will wrap to the first row of hearts. 
- if a player has more than 50 hearts, the silver heart render will move to the first row of hearts to prevent overlapping hearts.  Hearts will also adjust their Y-Position as more rows of hearts are added to the screen. 
- fixes silver hearts not bumping upwards correctly when regeneration is active.
- fixes silver hearts overlapping absorption hearts. 
- fixes silver hearts unrendering at the wrong time when you have more than 20 hearts. 

Heart overlay code sucks, don't ever touch it if you have a choice. :)

closes #822

---

### DIFF
--- a/src/main/java/com/gildedgames/aether/client/renderer/AetherOverlays.java
+++ b/src/main/java/com/gildedgames/aether/client/renderer/AetherOverlays.java
@@ -265,7 +265,7 @@ public class AetherOverlays {
                     int maxDefaultHealth = Mth.ceil(overallHealth - maxLifeShardHealth);
 
                     int currentOverallHealth = Mth.ceil(player.getHealth());
-                    int currentLifeShardHealth = Mth.ceil(player.getHealth() - maxDefaultHealth);
+                    int currentLifeShardHealth = Mth.ceil(maxDefaultHealth > 20 ? Mth.clamp(currentOverallHealth - 20, 0, maxLifeShardHealth) : currentOverallHealth - maxDefaultHealth);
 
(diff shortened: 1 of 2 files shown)


bconlon1 (approved):
lgtm

PR #878, opened 2022-12-03
fix(entity): Fix NPE crash when the slider collides with an entity

Burning-Cactus:
The slider now checks to see if the goal exists before calling stop().


bconlon1 (approved):
lgtm

PR #882, opened 2022-12-06
fix(structure): Fix bronze dungeons preventing trees from generating

Burning-Cactus:
This PR prevents the `StructureStart` from inflating the bronze dungeon's bounding box.


bconlon1 (approved):
lgtm

PR #884, opened 2022-12-20
feat(structure): Random layouts for the silver dungeon

Burning-Cactus:
This PR improves upon the old algorithm to use structure templates for a randomized silver dungeon. Additionally, the valkyrie queen should properly detect the bounds of the whole dungeon.


bconlon1 (changes requested):
- Silver dungeons should only be able to have a maximum of one chest or chest mimic in the center of a room.
- Trapped blocks shouldn't spawn on walls.
- Angelic trapped blocks and also probably hellfire trapped blocks should be made to convert into locked block variants in their constructor instead of the unlocked default variants.

bconlon1 (changes requested):
There seem to only be Trapped Light Angelic Stone blocks spawning in the Silver Dungeon and not regular Trapped Angelic Stone as well.

bconlon1 (approved):
lgtm

PR #885, opened 2022-12-24
fix(item): Accessories no longer have "curio" tag

bconlon1:
This stops the generic accessories from being able to be put in every accessory slot


Burning-Cactus (approved):
Looks good to me!

PR #886, opened 2022-12-25
fix(block): Trapped angelic blocks are now unbreakable and hellfire blocks follow this pattern

bconlon1:


Burning-Cactus (approved):
Looks good to me!

PR #897, opened 2023-01-07
fix(entity): Fix the valkyrie AI

### LINKED ISSUE #776: Enhancement: Boss balancing and fixes

Burning-Cactus:
This PR should take care of any remaining issues with the bosses in the Aether.
Closes #776


bconlon1 (approved):
lgtm

Issue #274, opened 2021-12-18
Enhancement: Finalize dungeon structures


bconlon1:
General:
- [x] Distribution balancing; silver and gold might be too low in y-range on average, and need to be further apart. ratio is probably good though.
- [x] /locate breaks dungeon spawning.
---
Bronze Dungeon:
- [x] Exclusion zone for aerial dungeons.
- [x] Bronze Dungeon needs terrain shaping implemented with new world generation system. Or just better location checking.
- [x] Bronze dungeon edges can occasionally generate slightly out of terrain.
- [x] Bronze dungeons need more spacing; sometimes a tunnel can collide with another bronze dungeon (even if it doesn't do anything).
- [x] Potentially look into random placement of trapped sentry blocks instead of static placement.
---
Silver Dungeon:
- [x] Silver dungeon aerclouds should be double drops.
- [x] Silver dungeon front shouldn't be able to be blocked by terrain; change it to start from the front of the dungeon instead of the back.
- [x] Stairs are using void not air.
- [x] Silver dungeon exterior shouldnt replace blocks; edges should be void instead of air so they don't replace terrain.
- [x] Lower minimum y-level for Silver Dungeon spawning; should be doable once world generation changes make it so there are ways to guarantee they'll be in open space.
- [x] Fix Silver Dungeon performance stuttering and lag spikes.
- [x] Fix Silver Dungeon sometimes not spawning but the clouds do.
- [x] Fix Silver Dungeons seeming to be able to cut-off at chunk borders.
- [x] Implement randomized rooms for Silver Dungeons.
- [x] Stairs need to be adjusted so they dont do this:
![image](https://user-images.githubusercontent.com/67203206/210289668-d61b9803-8b3c-4de6-9e4b-93441f1fa7fc.png)
- [x] Adjust angelic light block distribution amount, there are some specific positions in the boss room that light blocks must spawn.
![image](https://user-images.githubusercontent.com/67203206/210288883-c702b466-0f53-48a1-9ea6-48dd5fc5773a.png)
![image](https://user-images.githubusercontent.com/67203206/210288893-735b30c1-9e93-4840-9956-bd3acc7effac.png)
---
Gold Dungeon:
- [x] Ensure entrance is always above ground,
- [x] Finalized golden oaks.
- [x] Lower minimum y-level for Gold  Dungeon spawning; should be doable once world generation changes make it so there are ways to guarantee they'll be in open space.
- [x] Whatever happened here:
[...]

PR #898, opened 2023-01-07
fix(structure): Fix inconsistencies in structures that were not present in the original versions

Burning-Cactus:
This PR will address several issues in #274.


bconlon1 (approved):
lgtm

PR #900, opened 2023-01-11
refactor: Address lack of wildcards for public API

Jonathing:
# You can't escape me.

Hi! Supersedes #808.

## Maintaining Public API

The Aether mod is a very popular mod and many people will want to make add-on mods (including, but not limited to, Lost Content). Therefore, it is imperative that your public API (which isn't just the `api` package, but anything that can be reasonably interfaced with by a third-party) is as smooth to work with as possible.

### The Trouble with Generics

The issue with generic types is that they are very static for what their purpose actually is. When you make a method that has a method parameter `Block`, the accepted argument should be of type `Block` but it can also be of any implementation of it. This is not the case with nested types, as they are not interpreted at runtime (generic types are erased at compile time *except for method and class type arguments*). This means that if a method parameter is `Supplier<Block>`, the required argument can be of a subtype of `Supplier` but the generic type must be declared as `Block` and not of any other subtype. This leads to issues where the third party (or even yourself) has to resort to hacky and potentially dangerous type erasure through unchecked casting to get the results necessary.

### Our unfortunate solution

The solution, of course, is to use wildcards. Obviously wildcards don't need to be used for absolutely everything (especially if a class is final), but it eliminates underlying issue I have stated earlier, making interfacing with the public API cleaner and painless. I've included as many of these wildcards as I could in this PR, while omitting the cases where wildcards would be inappropriate or unnecessary (such as private API, final classes, and return types in some cases).

## tl;dr

please use wildcards for public API thanks now merge me bitch

### DIFF
--- a/src/main/java/com/gildedgames/aether/api/registers/MoaType.java
+++ b/src/main/java/com/gildedgames/aether/api/registers/MoaType.java
@@ -10,7 +10,7 @@ import java.util.function.Supplier;
 
 public class MoaType
 {
-    private final Supplier<Item> egg;
+    private final Supplier<? extends Item> egg;
     private final int maxJumps;
     private final float speed;
     private final ResourceLocation texture;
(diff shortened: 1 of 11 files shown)


bconlon1 (approved):
lgtm

Jonathing:
thanks i'll be back with more chicken wings next month

PR #904, opened 2023-01-12
feat(entity): Snowballs damage fire minions now

Burning-Cactus:
Snowballs now deal 3 damage to fire minions on hit.


bconlon1 (approved):
lgtm

PR #905, opened 2023-01-13
fix: Even more fixes to the Valkyrie Tools' range limiting

bconlon1:
Hopefully they finally work without missed cases


Burning-Cactus (approved):
Looks good to me!

PR #906, opened 2023-01-14
refactor: Change accessor names and formatting

bconlon1:


Burning-Cactus (approved):
Looks good to me!

Jonathing:
I don't understand what the intention behind this was. Mods that make accessor mixins for other mods do not conflict with yours if they have the same method signature.

bconlon1:
This was done after an edge-case was found where an accessor for a field ended up overwriting a method instead, creating an issue that was very hard to debug.

Jonathing:
I don't think that warrants every single interface method having the `aether$` prefix. Prefixes like those should normally be considered an edge case. It's not important, but I just thought it was strange.

bconlon1:
It's mainly just to keep things consistent and avoid another edge case appearing, since the original issue didn't exist when an accessor was first created but occurred later on.

Issue #311, opened 2021-12-30
Chore: Clean up code, Java conventions, and mapping conventions.


bconlon1:
- Keep methods in fairly consistent order.
- Update parameters and variable names, make sure they're camelcase.
- Cleanup unused imports (use IntelliJ's code analyze option).
- Fix any spacing, unnecessary returns (also as in spacing), bracket placement, lack of usage of `this.`
- Use getters and setters and keep fields private for new fields.
- Try to use getters and setters for vanilla fields regardless of protection.
- Any usage of `@SupressWarnings` should probably have documentation detailing why they're there if more are put in. 
- Properly annotate with `@Override`.
- Look for ways to abstract code into functional interfaces in instances of repeated code.
- Abstract some worldgen (trees) and rendering stuff (multi model renderers) to Nitrogen

bconlon1:
Refactoring to `core.mixin` pending/in-progress with #310.

bconlon1:
`core.registry` and `core.api` are good.

bconlon1:
Refactoring to `core.capability`, `core.network`, `core.util` (except BlockPlacers), and `core.data` (except AetherWorldProvider) pending with #316.

bconlon1:
`core` package should be reviewed again for certain mapping-related refactors that have to be done still.

bconlon1:
~~Currently blocked by #518.~~

bconlon1:
Progress:
- [x] `advancement`
- [x] `api`
- [x] `block`
- [x] `blockentity`
- [x] `capability`
- [x] `client`
- [x] `command`
- [x] `data`
- [x] `effect`
- [x] `entity`
- [x] `event`
- [x] `integration` 
- [x] `inventory`
- [x] `item`
- [x] `loot`
- [x] `mixin`
- [x] `network`
- [x] `perk`
- [x] `recipe`
- [x] ~~`util`~~
- [x] `world` 
- [x] Aether
- [x] AetherConfig
- [x] AetherTags

bconlon1:
~~Block and Item refactoring pending on https://github.com/MinecraftForge/MinecraftForge/pull/8838.~~

bconlon1:
Various item/block fixes and refactors that need doing or looking into:
- [x]  Use more `<br>`s for javadoc readability. 
- [x] More capitalization for items in javadocs 
- [x] Make sure param javadocs link class.
- [x] Abstract full swing check to EquipmentUtil.
[...]

PR #908, opened 2023-01-17
fix(entity): Swets can grab crouching players. Frogs no longer eat swets with passengers.

Burning-Cactus:
See #311 

The new MountableMob interface is used for the shared steering code between swets and the other mountable mobs.


bconlon1 (approved):
lgtm

PR #910, opened 2023-01-17
fix(entity): Mimics now attack as soon as they find a player after spawning.

Burning-Cactus:
See #311 

ContinuousAttackGoal now uses the move control if there isn't a path. This is to make the goal less performance-intensive.


bconlon1 (approved):
lgtm

PR #911, opened 2023-01-17
fix(entity): The slider ignores shields

Burning-Cactus:
See #311 

The slider now ignores shields to prevent eating through their durability.


bconlon1 (approved):
lgtm

PR #913, opened 2023-01-23
feat: World gen improvements 

FlashFyreMods:
This PR:
- Adds density function and noise registry
- Adjusts the terrain gen to make it similar to the original terrain
- Moves silver and gold dungeons into a shared structure set
- Fixes and adds to the biome map for better biome distribution

---


bconlon1 (approved):
lgtm

PR #914, opened 2023-01-24
refactor: Balance feature generation

### LINKED ISSUE #272: Enhancement: Balance feature generation rates.
- [x] Make features utilize relative height scaling.
- [x] Placement modifier order matters; make sure all features are reflective of vanilla ordering to make sure they're correct.
- [x] Aerclouds
- [x] Large Aerclouds
- [x] Crystal Trees
- [x] Holiday Trees
- [x] Flowers (separate bushes from flowers)
- [x] Grass.
- [x] Bushes need to have lower densities in clusters
- [x] Quicksoil.
- [x] Lakes and springs need to be more common.
- [x] Fix trees spawning over lakes.
- [x] Ores
- [x] Ensure configurations and placements are both all accurate.
- [x] Balance between b1.7.3 and 1.2.5

### LINKED ISSUE #512: Enhancement: Balance biome and tree distribution.
- Tree distributions need reducing.
- Biomes might also need to have their rarities adjusted in general, but that might depend on how messing with tree distribution looks first. It feels like atm skyroot thickets should be less common and the average biome should be forest.

bconlon1:
- [x] closes #272 
- [x] closes #512


OzPayn (approved):
looks good to me!

PR #918, opened 2023-01-26
fix(entity): Fix leftover bugs for swets and valkyrie queens

Burning-Cactus:
Swets no longer spawn slime particles or use slime sound events.
Valkyrie queens shouldn't teleport outside the room anymore.


bconlon1 (approved):
lgtm

PR #919, opened 2023-01-26
fix: Glove and invisibility fixes

bconlon1:
- [x] Gloves render in first person even when using an invisibility cloak.
- [x] Gloves don't render in first person when having the invisibility status effect even when they do in third person.
- [x] Taking off invisibility cloak brings back armor rendering first and then player rendering; there's a noticeable delay between the two.
- [x] Taking off invisiblity cloak if another invisibility effect is active removes the invisibility from that as well.


Burning-Cactus (approved):
Looks good to me!

PR #921, opened 2023-01-27
refactor: Tweak ore gen

bconlon1:


OzPayn (approved):
tested a bunch, seems good

PR #924, opened 2023-01-29
fix(entity): Fix aerbunny mount bugs

bconlon1:
Aerbunnies will now be dismounted if the player is in spectator, and other players can no longer interact with an Aerbunny that is already mounted


Burning-Cactus (approved):
Looks good to me!

PR #926, opened 2023-01-29
fix(item): Fix overlay for Hammer of Notch and also rebalance

bconlon1:
Reduces cooldown and increases the range of projectile knockback area


Burning-Cactus (approved):
Looks good to me!

PR #929, opened 2023-01-29
refactor: Golden Oaks generate more similar to how they used to

reetamb:


OzPayn (approved):
yep looks good to me!

PR #932, opened 2023-01-29
fix: Equip sound fixes for armor

bconlon1:


OzPayn (approved):
looks good (accessory sfx still missing but pending curios fix)

PR #941, opened 2023-02-04
fix: Bosses can no longer be pushed through doorways at all

bconlon1:
Doorway blocks will behave as solid towards their respective boss mobs


reetamb (approved):
i like this one its a funny issue to have to fix

PR #943, opened 2023-02-04
fix: Book of Lore resets pages properly and lore is more accurate

bconlon1:


OzPayn (approved):
lgtm, lore changes seem good

PR #944, opened 2023-02-07
feat(structure): Structure placements

### LINKED ISSUE #274: Enhancement: Finalize dungeon structures

Burning-Cactus:
This PR will make structures generate at better frequencies for gameplay, and hopefully make them a little more interesting.

See #274


OzPayn (approved):
Looks good so far, will add my own rebalances in a separate PR when they're more finalised

PR #945, opened 2023-02-07
refactor: Make lava freezing quieter

bconlon1:


OzPayn (approved):
yep that works, good stuff

PR #949, opened 2023-02-10
refactor(block): Experiment with original-accurate block hardness

bconlon1:


OzPayn (approved):
All seems accurate but we should get some extra testing to make sure post-merge

PR #951, opened 2023-02-10
fix(portal): Prevent portals from spawning on aerclouds

Burning-Cactus:


bconlon1 (changes requested):
Maybe we should just add a single block tag for all blocks that portals can't be spawned on so this list doesn't keep increasing and so that there's better compatibility for addons.

bconlon1 (approved):
lgtm

bconlon1 (changes requested):
Only have tag contain blue aerclouds

bconlon1 (approved):
lgtm

PR #954, opened 2023-02-11
adjusted ore distribution

OzPayn:
-raised gravitite height to 50
-raised zanite height to 75
-raised gravitite rarity
-lowered gravitite vein size

should feel better now, helps with some higher islands


devilbringer (approved):
Tested ore gen values and feels good.

bconlon1 (approved):
lgtm

PR #957, opened 2023-02-11
fix: Rebalance spawn rates

Burning-Cactus:


OzPayn (approved):
yep these feel much better good job

PR #968, opened 2023-02-15
refactor: More mob spawn changes

Burning-Cactus:


OzPayn (approved):
looks good to me, can't wait for ANTS players to complain and we have to do this again

PR #969, opened 2023-02-15
Update phyg.png

OzPayn:
Phyg was not consistent with the original texture so I fixed it with a quick palette swap


OzPayn:
image
Phyg Restored

bconlon1 (approved):
lgtm

PR #970, opened 2023-02-15
rebalanced most food items

OzPayn:
most foods now have better saturation, gummy swets are slightly debuffed, enchanted berries have less nutrition but much much higher saturation. 
Regeneration from eating should feel much better now


devilbringer (approved):
Tested new food values, feels far better.

PR #971, opened 2023-02-16
refactor: Spaced out floating structures

OzPayn:
Expanded separation between floating dungeons, so large structures don't spawn near each other any more. Enlarged the spacing distance too to allow for more variance in spawn locations.

image
Max render distance view of another close by dungeon for a sense of scale.


bconlon1 (approved):
lgtm

PR #972, opened 2023-02-16
fix: slider top layer emissive

OzPayn:
the red glow was way blown out so I edited it to display more correctly

image
before

image
after


bconlon1 (approved):
lgtm

devilbringer (approved):
Tested, looks much better.

PR #976, opened 2023-02-17
fix: Gravitite-related exploits

bconlon1:
- Gravitite Ore can no longer be harvested by using a torch if the ore is natural; trying to break it will instead break the torch; non-natural floating blocks like powered Enchanted Gravitite and ones from Gravitite Tools will break when interacting with these kinds of blocks though. 
- Enchanted Gravitite has a better placement check to prevent a duplication bug.


OzPayn:
So the fallback for this is just the block gets deleted and nothing drops? Just want to make sure this is what we're going with (should be fine given it's a user-initiated action but it is *quite destructive*)

OzPayn:
I feel like the block should replace the torch, like it should destroy it as it passes through it similarly to water

OzPayn:
Just tried it out, works perfectly!

OzPayn (approved):
lgtm

PR #977, opened 2023-02-18
Improved Moa mount functionality

OzPayn:
Jump values have been buffed, glide speed has been buffed. Should feel a lot better for long range horizontal movement and also makes the power difference between Blue, White and Black Moas for vertical movement more significant.


devilbringer (approved):
New values feel powerful and the difference between tiers is significant. Moas now feel like the stand out traversal mechanic in the aether, especially after recent aerbunny nerf.

PR #978, opened 2023-02-18
nerfed swets

OzPayn:
Reduced a lot of the Swet attributes so they better reflect their intended level of annoyance and difficulty within the mod's design sandbox. Should be a lot easier to deal with going forward.


devilbringer (approved):
So much better good lord thank you.

PR #979, opened 2023-02-18
mild aerbunny debuff

OzPayn:
reduced the rate at which you can activate extra mid-air jumps so the aerbunny fits into the traversal progression better


devilbringer (approved):
Reduced jump rate feels far better.

PR #980, opened 2023-02-18
shield recipe additions

OzPayn:
-added new recipes that allow skyroot-zanite and vanilla wood-zanite shield crafting

-changed the recipe names to make addon conflicts less likely


devilbringer (approved):
All recipes work as intended an unlock in the guide upon obtaining the correct relevant item.

PR #983, opened 2023-02-19
add more aether block convert recipe support on jei

baguchi:
This pull request make more jei compat for aether

What Add it on Jei Recipe Support:

Swet Ball Convation 
Ambrosium Convation

---


src/main/java/com/gildedgames/aether/data/generators/AetherLanguageData.java
```diff
@@ -622,6 +622,8 @@ protected void addTranslations() {
         addGuiText("jei.freezing", "Freezing");
         addGuiText("jei.incubating", "Incubating");
         addGuiText("jei.fuel", "Aether Fuel");
+        addGuiText("jei.swet_ball_convert", "Swet Ball Convert");
+        addGuiText("jei.ambrosium_convert", "Ambrosium Convert");
```
Jonathing: English isn't your mother tongue, which is perfectly fine. I still want to be picky about it though for the sake of consistency, the PR, and also to help you practice.

Instead of "Ambrosium Convert", "Ambrosium Conversion" would be better, since the "-sion" part of the word implies an effect, which is what these JEI recipes explain. Same for "Swet Ball Convert".

Jonathing (commented):
Good PR. Short and simple. Please do take note about my comment on the English translation keys for swet ball conversion and ambrosium conversion.

Both swet ball conversion and ambrosium conversion are so simple, that their GUI classes for each look very similar (or maybe even are the same). It might be prudent to make a superclass or super interface to reduce the boilerplate when adding more JEI recipes like these. For example, interface `ConversionRecipe` could be interfaced in both `AmbrosiumRecipe` and `SwetBallRecipe` to alleviate this potential problem. Though, that would probably be a job for @bconlon1 or a PR from myself if it came down to it.

baguchi:
Thanks for kindly help :)
yeah. It's true that the internals are similar, so it might be possible to omit it.

OzPayn (approved):
lgtm

PR #984, opened 2023-02-20
Fix Silver Dungeons' entrances generating inside terrain

Drullkus:
Shift Silver Dungeons' center of generation to terrain heightmap sampling position, for ensuring entrance always stays above ground


bconlon1 (approved):
lgtm

PR #986, opened 2023-02-20
fix: Bronze Dungeon generation avoids overlapping airborne dungeons

Drullkus:
Bronze Dungeon generation avoid overlapping airborne dungeons, especially for cases when said airborne dungeon has also generated through the terrain


bconlon1 (approved):
lgtm

PR #988, opened 2023-02-20
improv: Trapped blocks

Burning-Cactus:
Trapped blocks now attempt to spawn mobs 3 blocks away from the player.
Trapped blocks are quieter.


OzPayn (approved):
Works really well, good stuff

PR #991, opened 2023-02-20
refactor: Subtitles use aether namespace and crafters have unique subtitles

bconlon1:


OzPayn:
The new crafters still use the Furnace Crackle subtitle, otherwise everything looks good

OzPayn (approved):
yep looks good

PR #992, opened 2023-02-20
fix: Moa render crashing if a Moa is rendered before its type is set

Drullkus:


Jonathing:
I think it would be worth trying to somehow enforce the user (developer) to set the moa type upon the moa's creation before it is freshly added to the world. This is a good workaround in case something goes horribly wrong, though.

Drullkus:
The real crash this PR fixes is when the client has not received Moa’s information in time for it to begin rendering. It’s basically a race condition and this PR eliminates the root cause.

I would’ve actually just short-circuited the main render method for it to render nothing if the type were null, but I settled for just passing a default texture instead instead of writing a render method override

bconlon1 (approved):
lgtm

Jonathing:
Ah, ok. Was wondering what the issue would've been in that case. A client desync sounds like a normal case to me.

PR #993, opened 2023-02-21
fix(entity): Invisbility no longer confuses Slider

bconlon1:


devilbringer (approved):
DID YOU THINK YOU COULD HIDE?!

PR #997, opened 2023-02-21
Mob rebalancing (Deluxe edition)

OzPayn:
Zephyr behaviour has been reworked to be less aggressive
Mob spawning has been balanced to account for nighttime
Zephyrs are also generally less common


devilbringer (approved):
Zephyrs now feel present and active yet not overwhelmingly hostile/persistent, much better overall early game feel.
Swets are far less annoying.
Nighttime mob spawning no longer overkill.

PR #998, opened 2023-02-21
Dungeon Spawning Rebalancing

OzPayn:


devilbringer (approved):
Bronze dungeons are common but rarely group up, when they do they tend not to intersect much if at all.
Silver and Gold dungeons spawn regularly but more rarely now, feels significant when you find one now.

PR #1004, opened 2023-02-23
fix: Remove orange moas

Burning-Cactus:
Orange moas do not fit in with Aether 1 in its current state. They fit more in an addon mod.

If the game tries to load an already existing orange moa, there will be an error in the log. This'll only happen once per moa, so it should be fine.


bconlon1 (approved):
lgtm

PR #1005, opened 2023-02-23
build: Add Debug Utils as a runtime dependency

Burning-Cactus:
https://www.curseforge.com/minecraft/mc-mods/debug-utils-forge


bconlon1 (approved):
lgtm

PR #1006, opened 2023-02-23
feat: Adjust lightmap to not be warm-tinted

bconlon1:
Now:

Previous:


OzPayn:
Works well but needs to be an off-by-default config rather than replacing the current lightmap. Looks good for b1.7 parity but doesn't match the updated aesthetic from 1.2 onwards.

OzPayn (approved):
lgtm

PR #1008, opened 2023-02-24
fix: Parachute can no longer auto-deploy when player already has a vehicle

### LINKED ISSUE #1007: Bug: Crouching while on a Moa can activate Parachutes automatically

Entity

### Other Type

_No response_

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

44.1.17

### The Aether Version

0.3.1

no

### Client Log

_No response_

_No response_

### Steps to Reproduce

Fly up high with a Moa with a Gold parachute in inventory
Start falling
Hold crouch to begin fast falling
Eventually (if high enough/fast enough) the parachute will deploy and kick you off your Moa
Usually because you're holding crouch you will also dismount the parachute and fall

### What You Expect To Happen

The Moa falls and the parachute does not activate

### What Actually Happened

The Moa falls and the parachute activates

### Additional Details

This is annoying

bconlon1:
closes #1007


OzPayn (approved):
lgtm

PR #1012, opened 2023-02-26
feat: Use entity processing to make bosses for dungeons

Burning-Cactus:
This fixes a mod conflict with Distant Horizons and gives data packers a new tool to play with.


OzPayn (approved):
lgtm

PR #1016, opened 2023-02-27
improved aercloud aesthetics

OzPayn:
increased opacity on new textures
added shadow casting
added ambient occlusion

Aerclouds now closer resemble the rendering in the original mod, higher opacity hides opposite faces better and shadows create a softer, less glassy look.

1.2
image

this PR
image


bconlon1 (approved):
lgtm

PR #1019, opened 2023-02-28
feat: Green sunset config

bconlon1:


OzPayn (approved):
green colour needs adjusting at some point but this is good/accurate enough for now, and the config works fine

PR #1033, opened 2023-03-05
fix: Pressing the Aether inventory key now closes containers

Burning-Cactus:


OzPayn:
Pressing the aether inventory button currently closes the chat

OzPayn:
As we discussed before I think this should close containers but not immediately reopen the Aether inventory, similar to the vanilla inventory key

PR #1035, opened 2023-03-05
fix: rebalanced fuel and timing for utility blocks

OzPayn:
Altar repair times shortened
Altar enchanting times shortened
Ambrosium fuel time reduced slightly
Incubation time cut drastically
Freezer timings mildly rebalanced


Burning-Cactus (approved):
Looks good to me!

PR #1038, opened 2023-03-07
feat: added swet ball recipes for remaining slimeball recipes

OzPayn:
added swet slime block and swet sticky piston recipes


Burning-Cactus (approved):
Looks good to me!

PR #1039, opened 2023-03-08
fix: added b1.7 zanite and gravitite block textures

OzPayn:
moved the 1.2 textures from the base pack into the 1.2 pack and added recreated b1.7 textures to the b1.7 pack

image
beta 1.7 resources


devilbringer (approved):
horrible, perfect.

PR #1041, opened 2023-03-08
fix: The sun spirit no longer despawns on peaceful

Burning-Cactus:


OzPayn (approved):
he's not very peaceful :(
good job!

PR #1065, opened 2023-03-13
refactor: Reworked phoenix armor and neptune armor set logic

### LINKED ISSUE #1064: Bug: Phoenix armor struggles against lava currents

Item

### Other Type

_No response_

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

44.1.0

### The Aether Version

1.19.3-0.4.0

_No response_

### Client Log

_No response_

_No response_

### Steps to Reproduce

Seems to have a noticeable effect in Nether only.

1. Go to the Nether.
2. Dive into the lava ocean while wearing full phoenix armor.
3. Dig 2 blocks down when standing on the bottom.
4. Try to swim out.

Or

1. Go to the Nether.
2. Spill lava on a flat area.
3. Try to reach the source block while in phoenix armor.

### What You Expect To Happen

I expect to have good control of my character at all times while touching lava.

### What Actually Happened

My character is unable to swim out of a hole with the lava dragging it down and cannot push against horizontally flowing lava. Both actions are possible when a piece of armor is taken off.

### Additional Details

I was unable to reproduce this issue in the Overworld, the currents were never strong enough.
Neptune armor seems to have similar issues when in water (gets dragged down / pushed faster and gets pulled by magma block bubbles faster), but the impact is far less noticeable and does not create situations where taking it off is necessary.

Burning-Cactus:
Phoenix armor works based off of acceleration instead of the entity's current speed. Neptune armor uses attributes.

Closes #1064


Cinex20:
Phoenix armor seems to work fine now, though the speed with depth strider is much higher than before. Neptune armor, on the other hand, stopped working completely.

Cinex20:
Speed with phoenix seems much more reasonable now. Neptune armor still doesn't work.

Burning-Cactus:
Neptune armor works fine, it's just more subtle.

OzPayn (approved):
Looks good to me, both sets feel like they make more sense. Phoenix ability now feels more consistent and Neptune not needing the full set for the swim speed effect to start is much better

PR #1081, opened 2023-03-20
fix: Parachute rendering and shadows

### LINKED ISSUE #1078: Bug: Parachute renders don't have shadows/shading

Burning-Cactus:
Closes #1078


bconlon1 (changes requested):
Adding the shadow doesn't fix the issue of the parachute face shading, maybe changing the renderer to work like https://github.com/Gilded-Games/The-Aether/blob/1.19/src/main/java/com/gildedgames/aether/client/renderer/entity/FloatingBlockRenderer.java would fix the issue?

Drullkus (approved):
Looking fixed!

bconlon1 (approved):
lgtm

PR #1083, opened 2023-03-22
fix: Remove workaround for milking that may have been causing issues

### LINKED ISSUE #1043: Bug: Skyroot buckets can be duped by milking cows

Item

### Other Type

_No response_

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

44.1.0

### The Aether Version

1.19.3-0.4.0

_No response_

### Client Log

_No response_

_No response_

### Steps to Reproduce

1. Grab a few skyroot buckets (I needed at least 3 for it to work)
2. Spam-click on a cow.

### What You Expect To Happen

I expected to receive as many buckets of milk as I was holding in my hand.

### What Actually Happened

I received one more bucket than I was supposed to.

### Additional Details

I was not able to reproduce this bug on a singleplayer world, only ANTS.
Another player on the server was unable to reproduce this bug.
Attempting to perform the dupe using a body of water instead of milk did not work.
Iron buckets seemed to be duplicable at first, but the duped items vanished when interacted with in any way.

bconlon1:
closes #1043


Drullkus (approved):
Looking good!

PR #1086, opened 2023-03-23
fix: End Portals no longer dupe floating blocks

### LINKED ISSUE #1074: Rising blocks accelerated by blue aerclouds duplicate themselves when entering an end portal obstructed by blocks on the other side
https://youtu.be/C7T_yghkcyQ

How to reproduce: Shown in video, but here is a brief text description: 

Cause any block that is not gravitite ore to rise, either via redstone or a gravitite tool. Before colliding with the end portal block, it must be accelerated via blue aercloud. The block just above the portal must not be passable space for the rising block, so it attempts to place itself on the end portal block. Upon entering the end, the rising block will have duplicated itself. 

Expected behavior: rising blocks teleport to the end and break if their path is obstructed.

Observed behavior: rising blocks teleport to the end and break at the same time, resulting in a duplication of the block.

bconlon1:
closes #1074


Drullkus (approved):
Dupe not reproducible!

PR #1088, opened 2023-03-23
fix: Better accessory shift click quick move functionality

### LINKED ISSUE #1067: Bug: Shift-clicking a curio that is already equipped moves it to the destroy item slot
How to reproduce:
1. Equip a pair of gloves
2. Attempt to shift-click another pair of gloves into the slot

bconlon1:
closes #1067


OzPayn (approved):
yep that's fixed that

PR #1091, opened 2023-03-24
Enhancement: Compressed panorama files

### LINKED ISSUE #1087: Enhancement: Try to compress panorama to reduce file size

OzPayn:
Only real visual artefact is slightly reduced colour depth and the gradient from the horizon being quite blocky (but doesn't look incorrect imo) file size reduction is about 20-30% per file

before
image

after
image

closes #1087


bconlon1 (approved):
lgtm

PR #1092, opened 2023-03-24
Enhancement: Slider behaviour experiments

OzPayn:
Slider behaviour changes including:
randomised timing between actions
health dependant speed
higher overall speed


bconlon1 (approved):
lgtm

PR #1120, opened 2023-04-06
fix(block): Correct slab hardness

### LINKED ISSUE #1098: Bug: Aether slabs break faster than they should

Block

### Other Type

_No response_

### What Type of Bug Is This?

None

### Forge Version

45.0.24

### The Aether Version

0.5.1

https://gist.github.com/kodadacod/f6c498a245d66b67d9937c9bd57f806f

### Client Log

_No response_

_No response_

### Steps to Reproduce

Everytime
1. Place down a slab and block
2. Break them both

### What You Expect To Happen

The slab and the block would share the same breaking time

### What Actually Happened

The slab took at least half the time to break than the full block

### Additional Details

https://streamable.com/yg23c7

bconlon1:
closes #1098


Burning-Cactus (approved):
Looks good to me!

PR #1148, opened 2023-04-16
build: Generalize the buildscript and make it use plugin DSL

Jonathing:
# oh god he's back

Your `build.gradle` file is pretty nice and clean, but I want to help streamline its usage in the future, considering how this is not the only mod you lot will be working on. Here's a list of changes and their justifications:

- Plugins are now applied using the Gradle plugin DSL. This solves the deprecated usage of applying plugins manually via the `apply plugin:` syntax. It also thins out buildscript repositories (as they are technically now located in `settings.gradle`) and can make plugin management easier to maintain.
- Removed the "eclipse" plugin. This plugin is deprecated and should not be used, as it has shown to cause problems with MixinGradle and even Eclipse itself. The proper way to use Eclipse with a project is to imported it with "Existing Gradle Project", and it will generate the `classpath` and `project` files for you. `genEclipseRuns` functions as always and running the game posed no problems.
- The mod ID and name are now defined in `gradle.properties`. This solves the re-usage of "aether" and "The Aether" in the buildscript and thus generalizes it for usage in other mods.
- Removed `MixinConfigs` from the jar manifest. MixinGradle already does this for you. What it wouldn't do for you are Mixin connectors, which you are not using.

All changes were tested by building a jar file with all artifacts intact, including Jar-in-Jar artifacts, the mixin refmap, the jar manifest including the `MixinConfigs` property, and Gradle not literally imploding when I tried to do anything else.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,0 +1,19 @@
+pluginManagement {
+    resolutionStrategy {
+        eachPlugin {
+            var plugin = requested.id.toString()
+
+            switch (plugin) {
+                case 'org.spongepowered.mixin':
+                    useModule "org.spongepowered:mixingradle:${requested.version}"
+                    break
+            }
+        }
+    }
+    repositories {
+        gradlePluginPortal()
+        maven { url = 'https://maven.minecraftforge.net/' }
+        maven { url = 'https://repo.spongepowered.org/maven' }
+        maven { url = 'https://maven.parchmentmc.org' }
+    }
+}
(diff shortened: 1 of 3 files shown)


settings.gradle
```diff
@@ -0,0 +1,19 @@
+pluginManagement {
+    resolutionStrategy {
+        eachPlugin {
+            var plugin = requested.id.toString()
+
+            switch (plugin) {
+                case 'org.spongepowered.mixin':
+                    useModule "org.spongepowered:mixingradle:${requested.version}"
+                    break
```
Jonathing, line 9: This syntax can be used with any plugin that does not have a plugin module in their repository, which in this case is just MixinGradle.

bconlon1 (approved):
lgtm

PR #1154, opened 2023-04-18
build: Isolate dependency management with GradleWorks

Jonathing:
## but I'm not done yet!

Big Disclaimer: I don't expect this PR to be merged. In fact, I prefer you decline this PR if not everyone on the dev team is on board with this. This is a proposal, not a contribution.

# GradleWorks

I've created a dependency management plugin for Gradle that isolates all dependency management into an external file called `dependencies.yaml`. The benefit of this approach is obvious: instead of needing to rely on manually editing dependencies in the middle of the `build.gradle` buildscript and making arbitrary variables in `gradle.properties` it is instead all controlled in `dependencies.yaml`. Additionally, the only thing this changes about the buildscript is dependency management. In the `minecraft`, `dependencies`, and `repositories` blocks, the methods `handleMinecraft()`, `handleDependencies()`, and `handleRepositories()` are called respectively, which all apply the information provided in the `dependencies.yaml` file. You can see how that file is structured in the PR.

As an important note, GradleWorks does not replace any code blocks from the build script. Look in the PR at those blocks to see what I mean. Example: If for some reason you have a special dependency that can't be put in the YAML file, you can write it in Groovy in the buildscript under the `gradleworks.handleDependencies()` line.

I've went ahead and copied every single dependency from `build.gradle` (with the notable exception of SmoothBoot for forge) and translated them to the YAML file. The YAML file itself is self-explanatory with what each dependency needs for it to function properly. GradleWorks also gives the option to enable or disable specific dependencies that might not be up-to-date or necessary (like JEI runtime, or the commented-out dependencies that were in the buildscript).

## Please Remember

This is less of a contribution and more of a proposal from me to use this new dependency management system. I've been using it myself in my own projects for a while now, and I believe I've developed it to a point where it truly has practical use in production.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ plugins {
     id 'io.github.0ffz.github-packages' version '1.+'
     id 'com.matthewprenger.cursegradle' version '1.4.+'
     id 'com.modrinth.minotaur' version '2.+'
-    id 'me.jonathing.gradleworks' version '6.0.0-beta.7'
+    id 'me.jonathing.gradleworks' version '6.0.0-beta.8'
 }
 
 version = "${gradleworks.minecraft().version}-${project.mod_version}-forge"
--- a/build.gradle
+++ b/build.gradle
@@ -81,7 +81,7 @@ repositories {
     gradleworks.handleRepositories()
 
     // special cases that gradleworks can't handle (like github packages) down here
-//    maven githubPackage.invoke("Gilded-Games/Nitrogen")Nitrogen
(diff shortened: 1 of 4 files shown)


Jonathing:
After a suitable amount of testing and jar building, I've marked this PR as ready for review. I am unable to test the CurseForge and Modrinth debug tests, because I don't have an API key to play with, so let me know if the JSONs the debug mods come up with look weird.

bconlon1:
We've discussed and have decided we prefer the current gradle dependency system, the suggestion is appreciated though.

Jonathing:
Noted. Cheers!

PR #1173, opened 2023-04-21
fix: Prevent Aerbunny from despawning after used by player

### LINKED ISSUE #1152: Bug: Aerbunnies despawn when they shouldn't.
Supposedly Aerbunnies despawn even after being used by players; they probably shouldn't.

Jonathing:
Fixes #1152.

Once an aerbunny is successfully mounted on a player, the persistence flag is marked to disallow natural despawning of the aerbunny.

One important caveat to note is that this implementation comes with the risk of creating an exploit that involves mounting multiple aerbunnies and preventing them all from despawning, causing server entity AI overload. Any suggestions to combat this are welcome.

### DIFF
--- a/src/main/java/com/aetherteam/aether/entity/passive/Aerbunny.java
+++ b/src/main/java/com/aetherteam/aether/entity/passive/Aerbunny.java
@@ -99,6 +99,7 @@ public class Aerbunny extends AetherAnimal {
         if (this.isOnGround() || !this.getFeetBlockState().isAir() || (this.getVehicle() != null && (this.getVehicle().isOnGround() || !this.getVehicle().getFeetBlockState().isAir()))) {
             this.lastPos = null;
         }
+        this.checkDespawn();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


bconlon1:
I'd considered this but the caveat had made me avoid it since I wasn't sure of a way to combat it, it's something that may need some discussion/brainstorming.

Jonathing:
I am more than welcome to convert this PR into an RFC discussion. It is something that should be discussed because it has the potential problem of breaking the gameplay loop (i.e. keeping an aerbunny for later use at a dungeon entrance).

bconlon1:
We can discuss it in #development-aether probably at some point when I don't have an impending exam.

Jonathing:
I, too, suffer from university Ls.

bconlon1:
After thinking about this more I think I also need to learn more about Minecraft's despawning system because I am finding conflicting information on how it's meant to work with animals like Aerbunnies.

Jonathing:
Oh god

bconlon1:
lmao

bconlon1:
We're likely gonna go with a different solution to this.

Jonathing:
Was surprised you didn't close it already.

PR #1182, opened 2023-04-23
fix: Replace %s in Valkyrie death message

### LINKED ISSUE #1181: Bug: Valkyrie Dialogue Error

Entity

### Other Type

_No response_

### What Type of Bug Is This?

Annoyance

### Forge Version

45.0.46

### The Aether Version

0.6.2

_No response_

### Client Log

https://gist.github.com/dev-slayer/31c8ccf87f655ecd30f37e5cde855bcc

_No response_

### Steps to Reproduce

This is a dialogue from when the Valkyrie kills **you**, it can be replicated, but it may take a few times for the death message to be the same.

### What You Expect To Happen

I expected it to say: "Maybe some day, SlayerDev... maybe some day."

### What Actually Happened

It outputed: "Maybe some day, %s... maybe some day."

### Additional Details

I assume the method of dialogue isn't similar to tellraw, but if it is, this may work instead.

[{"text":"Maybe some day, "},{"selector":"@p"},{"text":"... maybe some day."}]

bconlon1:
closes #1181

### DIFF
--- a/src/main/java/com/aetherteam/aether/entity/monster/dungeon/Valkyrie.java
+++ b/src/main/java/com/aetherteam/aether/entity/monster/dungeon/Valkyrie.java
@@ -4,6 +4,7 @@ import com.aetherteam.aether.client.AetherSoundEvents;
 import com.aetherteam.aether.item.AetherItems;
 
 import net.minecraft.network.chat.Component;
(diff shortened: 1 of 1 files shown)


OzPayn (approved):
looks good

PR #1185, opened 2023-04-23
fix: Improved check for Aerbunny jump resets

### LINKED ISSUE #1163: Bug: Aerbunnies will sometimes refuse to jump when the player bounces with an aercloud

Entity

### Other Type

_No response_

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

45.0.40

### The Aether Version

1.19.4-0.6.2

_No response_

### Client Log

_No response_

_No response_

### Steps to Reproduce

Happens rarely.
1. With an aerbunny on your head, bounce using an aercloud in a variety of ways.
2. Eventually the aerbunny will refuse to jump in the air.

### What You Expect To Happen

The aerbunny will jump if the player is in the air and holding space.

### What Actually Happened

The aerbunny occasionally stops responding to jump inputs after an aercloud bounce.

### Additional Details

This can also occasionally happen as a result of momentum provided by whirlwinds, which leads me to believe it isn't an aercloud-specific problem. Additionally, it seems to have something to do with when the momentum is applied in the aerbunny's jump timer, as if you can consistently trigger a momentum change at specific times, you can consistently replicate the bug.

bconlon1:
closes #1163

### DIFF
--- a/src/main/java/com/aetherteam/aether/entity/passive/Aerbunny.java
+++ b/src/main/java/com/aetherteam/aether/entity/passive/Aerbunny.java
@@ -96,7 +96,7 @@ public class Aerbunny extends AetherAnimal {
             this.setPuffiness(0);
         }
         this.handlePlayerInput();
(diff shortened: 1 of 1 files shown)


OzPayn (approved):
yep that seems to work

PR #1190, opened 2023-04-23
refactor: saddled mob cannot death in void

### LINKED ISSUE #1188: Feature Request: Saddled entities like Moas not dying to void

baguchi:
closed #1188 

This pullrequest make saddled mob cannot death when fall out from aether (I mean aether void)

also I have question should we also work another saddled mob?(like minecraft pig)


bconlon1 (approved):
lgtm

PR #1191, opened 2023-04-23
Tweek: moa has random attribute like horse

### LINKED ISSUE #1186: Moa walking speed tweak

Entity

### Other Type

_No response_

I think that moa walking speed should be tweaked to be faster.

with the upcoming patron bonuses, I think faster moas would be a good call as right now sprint jumping is faster than them on land and Elytra flight is way faster for moving around the aether. if they were given horse speed I think that there would be more of an incentive for players to use them mid-game and for patron skins.

_No response_

baguchi:
This pull request make moa has random attribute like horse
It will be able to increase the opportunities to use moa

here is change:
moa health is now 30 + and Generate a random number with a maximum of 5 twice and give it to moa (old is 35 health)
moa speed has 0.95F + random double * 0.25D (old is 1.0F)

### DIFF
--- a/src/main/java/com/aetherteam/aether/entity/passive/Moa.java
+++ b/src/main/java/com/aetherteam/aether/entity/passive/Moa.java
@@ -1,20 +1,16 @@
 package com.aetherteam.aether.entity.passive;
 
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
-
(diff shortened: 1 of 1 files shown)


bconlon1:
We'd prefer to keep the Moa stats fairly consistent so this falls outside our design goals.

PR #1254, opened 2023-04-29
feat: added Jade compat

GizmoTheMoonPig:
Adds compat for the mod [Jade](https://www.curseforge.com/minecraft/mc-mods/jade), which is an information HUD mod similar to WAILA or HWYLA. The following things have been adjusted:
- Chest mimic blocks will now display as normal chests
- Trapped dungeon blocks will now display as normal dungeon blocks
- both doorways will show up as locked dungeon blocks

---


bconlon1 (approved):
lgtm

PR #1292, opened 2023-05-04
feat: Tags for skyroot stick and tool crafting

### LINKED ISSUE #1287: Feature Request: Aether Planks Item Tag

_No response_

### Other Type

Tags/Addon-useable content

- [ ] I confirm this request is an enhancement to a feature already in The Aether.

An item tag that can be used to craft Skyroot tools and sticks, so that addon mods can add new items that craft into these without needing to add new recipes for every single item each time they add something new.

So that addons can more easily add alternatives to skyroot planks for crafting recipes

_No response_

bconlon1:
closes #1287


OzPayn (approved):
lgtm

PR #1298, opened 2023-05-04
fix: Zanite accessory durability won't be used by instamine blocks

### LINKED ISSUE #1294: Bug: Zanite accessories take damage when breaking instamineable blocks

_No response_

### Other Type

_No response_

### What Type of Bug Is This?

None

### Forge Version

45.0.50

### The Aether Version

1.19.4-1.0.0-beta.1

_No response_

### Client Log

_No response_

_No response_

### Steps to Reproduce

Always reproducible.
1. Equip a zanite ring or pendant.
2. Break an instamineable block (torch, grass, etc.)
3. Look at the accessories' durability.

### What You Expect To Happen

I expect the accessories to behave like tools and take no damage.

### What Actually Happened

The accessories' durability went down.

### Additional Details

_No response_

bconlon1:
closes #1294


OzPayn (approved):
looks like that's working

PR #1303, opened 2023-05-06
feat(entity): added lootr support to mimics

### LINKED ISSUE #1274: Feature Request: Lootr compat for Mimics

Entity

### Other Type

_No response_

- [ ] I confirm this request is an enhancement to a feature already in The Aether.

Make Mimics take on the Lootr Chest texture when Lootr is present. The mod Artifacts does this with their Mimics, so I assume it's possible to do here too.

As it stands, having Lootr present makes the Mimics stand out since they still look like normal chests while all real loot Chests get converted to Lootr's Chests

Lootr 1.19.4-0.6.26.63

GizmoTheMoonPig:
Mimics will now look like lootr chests if lootr is installed. This affects both the block and the entity.

closes #1274 

---


bconlon1 (approved):
lgtm

PR #1317, opened 2023-05-07
refactor!: Remove default Curios tags

### LINKED ISSUE #1304: "Duplicated" Curio slots

Item

### Other Type

_No response_

### What Type of Bug Is This?

Design

### Forge Version

45.0.50

### The Aether Version

1.0.0-beta.1

Curios API 5.1.5.3

### Client Log

_No response_

_No response_

### Steps to Reproduce

Always
1- Have both Curios API and Aether installed
2-Have 2 acessories (gloves for example)
3-You can equip both, one in the Curios acessory slot and other in the Ather Gloves slot

### What You Expect To Happen

Ather use/overwrite Curios Slots, allowing only one glove or other acessory at the same time.

### What Actually Happened

Aether is not using curios slots, its duplicating the slots, not using/overwritting or whatever is should do, letting me wear double the things (for example, 2 gloves)

### Additional Details

Word from on the discord report:

> bconlon — Today at 12:53 AM
> actually wait i just fully looked at the images i see whats going on here
> hmmmm
> this is a design problem i did not account for, report this to the issue tracker. i made it so accessories can work in other mods' curios slots as well but did not consider the fact that'd mean they can double up with ones other mods add 

bconlon1:
They seem to cause more issues than help when broadly applied for mod compat; this kind of compatibility will be left up to datapack creators

closes #1304


raptor494 (approved):
Maybe we could make a standalone curios compat datapack (downloaded separately?)

bconlon1:
> Maybe we could make a standalone curios compat datapack (downloaded separately?)

Yeah we can probably do something like that on release.

PR #1326, opened 2023-05-09
fix(translations): Fix German translation

Catchears:
sorted all keys according to `en_us.json`
added & translated missing keys
made words that have a vanilla translation (like 'Slab') use that vanilla translation 
standardized various terms that had multiple translations 
standardized the informal pronoun 'du' instead of formal 'sie'
improved clarity in some previously translated keys

---

I started this after seeing #1283, which - in all honesty, sorry to be harsh - raises quite a few question marks:

- ignores vanilla translations from the base game
- uses Äther and Aether interchangably
- includes json errors
- leaves single words in longer strings untranslated that already have translations
- duplicates some keys
- has inconsistencies in own changes (e.g. Verlies vs Dungeon) 
- out-of-context translations of some words, like translating 'drop' (from 'mob drops' or 'big drop') to Tropfen, which means drop of liquid)

---


bconlon1 (approved):
lgtm

PR #1344, opened 2023-05-11
feat: Golden Aercloud enchanting recipe

### LINKED ISSUE #1307: Feature Request: Golden Aercloud renewability with frogs

_No response_

### Other Type

_No response_

Currently, I don't believe there is a way to get aerclouds renewably, and I personally believe that there should at least be a way (feel free to disagree with me). I just noticed that frogs eat swets just like other slimes, so how about adding aercloud renewability by making frogs killing swets have a custom loot table?

Here's what I'm thinking. Make it so the white warm frogs drop cold aercloud when eating any two swet types since the white frogs are good at cooling off themselves in real life. Make the other two frogs drop only multiple blue aerclouds when eating blue swets and no swet balls, and make them drop golden aerclouds when eating golden swets instead of glowstone.

Personally I think it's kind of weird that aerclouds are not renewable. Maybe this is intended design but it just doesn't feel good considering golden and cold aerclouds are required to craft parachutes. It also prevents the player having to strip the aether of every cloud in sight, which would make the dimension look a little less pretty over time. Bringing frogs to the aether is also a relatively late game activity, so it theoretically shouldn't change early game balance that much.

_No response_

bconlon1:
closes #1307


OzPayn (approved):
yep that works for me

PR #1371, opened 2023-05-20
feat: JEI compat for in-world recipes

### LINKED ISSUE #852: Feature: JEI Compatibility (In-World Recipes)
JEI support for in-world recipes such as block banning.

May be more complex due to needing to display certain values like blockstates, biome tags if relevant, and may also require some additional GUI art or at least a GUI design mockup to plan the layout.

bconlon1:
closes #852


OzPayn (approved):
lgtm

PR #1386, opened 2023-05-20
fix: Aerclouds no longer act as ground during flight

### LINKED ISSUE #1380: Bug: Blue Aercloud canceling elytra flight

Block

### Other Type

_No response_

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

45.0.58

### The Aether Version

1.19.4-1.0.0-beta.3-forge

No

### Client Log

_No response_

_No response_

### Steps to Reproduce

Step 1) Deploy elytra
Step 2) Fly into blue aercloud

### What You Expect To Happen

Blue Aercloud effect to launch me up, and allow me to use the speed boost to extend flight range

### What Actually Happened

Elytra flight is cancelled, requiring me to redeploy the elytra and thus losing momentum

### Additional Details

_No response_

bconlon1:
closes #1380

### DIFF
--- a/src/main/java/com/aetherteam/aether/block/natural/AercloudBlock.java
+++ b/src/main/java/com/aetherteam/aether/block/natural/AercloudBlock.java
@@ -2,6 +2,7 @@ package com.aetherteam.aether.block.natural;
 
 import com.aetherteam.aether.block.AetherBlockStateProperties;
 import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.state.BlockState;
 import net.minecraft.world.level.block.HalfTransparentBlock;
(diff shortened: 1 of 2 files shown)


Jonathing (approved):

src/main/java/com/aetherteam/aether/block/natural/AercloudBlock.java
```diff
@@ -44,7 +45,9 @@ public void entityInside(BlockState state, Level level, BlockPos pos, Entity ent
 		if (entity.getDeltaMovement().y < 0.0) {
 			entity.setDeltaMovement(entity.getDeltaMovement().multiply(1.0, 0.005, 1.0));
 		}
-		entity.setOnGround(true);
+		if (!(entity instanceof Player player) || !player.getAbilities().flying) {
+			entity.setOnGround(true);
+		}
```
Jonathing, line 50: Actually, this does pose a slight problem: players who reach the bottom of the aercloud block while flying with an elytra will continue to "fly" in it, even though the block itself is not solid. Might it be better to only include this change in the blue aercloud?

OzPayn (approved):
works how I'd expect, gj

PR #1387, opened 2023-05-20
feat: Bucket tags for crafting

### LINKED ISSUE #1381: Feature Request: Bucket Tags

bconlon1:
closes #1381


Jonathing (commented):
This takes care of the bucket tags themselves, but there are no changes to recipes. It might be good to bundle Aether's recipes to change to using the tag in this PR, rather than just introduce the existence of these tags from the get-go.

bconlon1:
> This takes care of the bucket tags themselves, but there are no changes to recipes. It might be good to bundle Aether's recipes to change to using the tag in this PR, rather than just introduce the existence of these tags from the get-go.

That's because there are no Aether recipes that use Skyroot Milk Buckets to use this tag.

OzPayn (approved):
lgtm

PR #1389, opened 2023-05-20
improv!: Use mixins for capes for better compatibility

### LINKED ISSUE #1352: Bug: Duplicated cape rendering code

Item

### Other Type

_No response_

### What Type of Bug Is This?

Compatibility

### Forge Version

N/A

### The Aether Version

1.19.4-1.0.0-beta.2-forge

Essential 1.2.0.14

### Client Log

_No response_

_No response_

### Steps to Reproduce

1. Install The Aether and [Essential](https://essential.gg/) (also available via CF/Modrinth)
2. Enter world and equip any of Aether's capes
3. Open the Wardrobe from the Pause Menu
4. In the Emotes>Movement category, click the Running emote twice to unlock and equip it
5. Go back in-game
7. Press and hold R to open emote wheel
8. Click the running emote

### What You Expect To Happen

Cape should stay relatively attached to body and move around with it.

### What Actually Happened

Cape stays fixed in space (relate to entity position) completely independent of body movement.

(this is with Sleep emote; with Running it's not quite as obvious in a still image)

### Additional Details

This underlying issue here is that Aether seems to [copy the vanilla cape rendering code more or less verbatim](https://github.com/The-Aether-Team/The-Aether/blob/11a62f59e6cafaeea5de90d899e70a76439333c2/src/main/java/com/aetherteam/aether/client/renderer/accessory/CapeRenderer.java) and as such any third party modifications do not apply to it. This isn't really specific to Essential; I'd assume mods like Wavey Capes (if there's a forge equivalent) will also not work properly.

There's many more issues with simply coping large swaths of vanilla code. From it being difficult to tell what's different compared to vanilla, the code getting out of sync on updates, to the fact that copying non-trivial amounts of vanilla code is technically illegal.
So, unless there's something Aether does with its capes that I don't see (I'm not actually familiar with how they function ingame, I am just forwarding a report and am making educated guesses based on the code), I'd highly recommend simply replacing all that with a fairly simple mixin into `AbstractClientPlayerEntity.getCloakTextureLocation` (as done by [this mod](https://github.com/Kay9Unit/ItemCapes/blob/d1c894e76796ab0f0898ce61c9a1ba57533c15d2/src/main/java/com/github/kay9/itemcapes/mixins/client/AbstractClientPlayerEntityMixin.java) which does something fairly similar).

bconlon1:
closes #1352


src/main/java/com/aetherteam/aether/mixin/mixins/client/AbstractClientPlayerMixin.java
```diff
@@ -0,0 +1,37 @@
...
+        }
+    }
+
+    @Inject(at = @At("HEAD"), method = "getCloakTextureLocation", cancellable = true)
+    private void getCloakTextureLocation(CallbackInfoReturnable<ResourceLocation> cir) {
+        AbstractClientPlayer player = (AbstractClientPlayer) (Object) this;
+        SlotResult result = EquipmentUtil.getCape(player);
+        if (AetherMixinHooks.isCapeVisible(player) && result != null && result.stack().getItem() instanceof CapeItem capeItem) {
+            if (result.stack().getHoverName().getString().equalsIgnoreCase("swuff_'s cape")) {
+                cir.setReturnValue(new ResourceLocation(Aether.MODID, "textures/models/accessory/capes/swuff_accessory.png"));
```
Jonathing: I'd recommend storing this resource location in a global static variable.

OzPayn (approved):
tested with essential emotes, seems to work fine

PR #1404, opened 2023-05-22
feat: Better frosted ice and accessories only freeze source blocks

### LINKED ISSUE #1392: Bug: Frosted Ice doesn't unfreeze in the End
A vanilla bug that we probably have to work around by making our own frosted ice.

### LINKED ISSUE #1394: Bug: Ice Accessories freezing flowing liquids can cause issues
I think the accessories might not also have the proper ability to distinguish between flowing blocks and source blocks, I might just be checking for liquids and not blocks.

bconlon1:
closes #1394 
closes #1392


OzPayn:
With the frosted ice config disabled, freezing water source blocks doesn't seem to reliably stop the flowing water? Try placing a single water block in an open area then freezing it, only when you interact with that block or place blocks around it does the flowing water seem to stop? Kinda odd (doesn't happen with the frosted ice blocks)

OzPayn (approved):
yep that fixed it

PR #1427, opened 2023-05-24
fix: Sleep based on config.

bconlon1:


src/main/java/com/aetherteam/aether/event/listeners/DimensionListener.java
```diff
@@ -163,7 +164,7 @@ public static void onTriedToSleep(SleepingTimeCheckEvent event) {
         Player player = event.getEntity();
         if (player.getLevel().dimensionType().effectsLocation().equals(AetherDimensions.AETHER_DIMENSION_TYPE.location())) {
             AetherTime.get(player.getLevel()).ifPresent((aetherTime) -> {
-                if (aetherTime.getEternalDay()) {
+                if (aetherTime.getEternalDay() && !AetherConfig.COMMON.disable_eternal_day.get()) {
```
Jonathing: Dude, what is this? If eternal day is disabled in config, shouldn't `getEternalDay()` also return false?

Jonathing (approved):
Thank you.

Drullkus (approved):

PR #1433, opened 2023-05-25
improv!: Switch to server config for most configs

### LINKED ISSUE #1391: Bug: Some common configuration should be moved to server config
Common config exists for global values. Server config exists per-world. A lot of common config in the Aether really should be put in server config.

Here's the community wiki page for ForgeConfigSpec: <https://forge.gemwire.uk/wiki/Configs>
go nuts

[](https://discord.com/channels/223938153260515328/374854898879299585/1109850307409752144)

bconlon1:
This PR also removed ConfigCondition for recipes because switching to server configs means configs are not loaded before recipes, so recipes can't know config information. We can just make an official datapack for temporary ice accessory conversion on release.

closes #1391


Jonathing (changes requested):
Server config looks good. Thankfully since The Aether is still in beta, you can pull the "beta" card and say "well, we're in a beta so you should've expected breaking changes!! ! ! 😞" or some shit like that.

I do have a real gripe with this PR at its current state, though, and that is there are a couple of classes that have unnecessary logic changes for the scope of it. Shit like changes to recipe data generation and removal of `ConfigCondition`. It's probably important that you state those changes as well, since they made this way into this otherwise innocuous PR that all it should be doing is moving some configs over from common to server.

tl;dr server configs look nice, state all changes please

Jonathing (approved):

OzPayn (approved):
lgtm

PR #1443, opened 2023-05-27
feat: Moa Egg event

### LINKED ISSUE #1440: Feature Request: Moa Lay Egg event

Entity

### Other Type

_No response_

An event for when a moa lays an egg, that can cancel it or change the item, and that gives access to the moa

So that addons can make moas lay different items depending on the circumstances (for instance I would like to be able to make inebriated [or potentially some other effect I may add] moas lay a cockatrice egg instead of a moa egg)

_No response_

bconlon1:
closes #1440

### DIFF
--- a/src/main/java/com/aetherteam/aether/entity/passive/Moa.java
+++ b/src/main/java/com/aetherteam/aether/entity/passive/Moa.java
@@ -191,10 +191,14 @@ public class Moa extends MountableAnimal implements WingedBird {
 			if (!this.isBaby() && this.getPassengers().isEmpty() && --this.eggTime <= 0) {
 				MoaType moaType = this.getMoaType();
 				if (moaType != null) {
-					EggLayEvent eggLayEvent = AetherEventDispatch.onLayEgg(this, this.getMoaType().getEgg());
+					EggLayEvent eggLayEvent = AetherEventDispatch.onLayEgg(this, AetherSoundEvents.ENTITY_MOA_EGG.get(), 1.0F, (this.random.nextFloat() - this.random.nextFloat()) * 0.2F + 1.0F, this.getMoaType().getEgg());
 					if (!eggLayEvent.isCanceled()) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/aetherteam/aether/entity/passive/Moa.java
```diff
@@ -189,8 +191,11 @@ public void tick() {
 			if (!this.isBaby() && this.getPassengers().isEmpty() && --this.eggTime <= 0) {
 				MoaType moaType = this.getMoaType();
 				if (moaType != null) {
-					this.playSound(AetherSoundEvents.ENTITY_MOA_EGG.get(), 1.0F, (this.random.nextFloat() - this.random.nextFloat()) * 0.2F + 1.0F);
-					this.spawnAtLocation(this.getMoaType().getEgg());
+					EggLayEvent eggLayEvent = AetherEventDispatch.onLayEgg(this, this.getMoaType().getEgg());
+					if (!eggLayEvent.isCanceled()) {
+						this.playSound(AetherSoundEvents.ENTITY_MOA_EGG.get(), 1.0F, (this.random.nextFloat() - this.random.nextFloat()) * 0.2F + 1.0F);
```
Jonathing: There's so much more you can do with the `EggLayEvent`, including setting the sound event, sound volume, and sound pitch in the event, then using those values to play the sound. You can check for a null sound event and if it is null, just don't play a sound. It's a bit of a niche case, but it will be nice for it to exist in case add-on creators want to modify existing behavior without resorting to coremodding.

src/main/java/com/aetherteam/aether/event/events/EggLayEvent.java
```diff
@@ -0,0 +1,24 @@
...
+        this.item = item;
+    }
+
+    public Item getItem() {
+        return this.item;
+    }
+
+    public void setItem(Item item) {
+        this.item = item;
+    }
```
Jonathing: Since your repository isn't consistent about what should or shouldn't be nullable, I think it's important to set an annotation here or a Javadoc comment clarifying what it can be.

Jonathing (changes requested):

Jonathing (approved):

reetamb (approved):

PR #1449, opened 2023-05-27
refactor!: Remove slim and non-sleeve gloves

### LINKED ISSUE #1441: Enhancement: Remove slim gloves
Slim gloves are likely to cause a lot of headaches and issues if we want to add trims in 1.20 to gloves. They make texturing armour tedious, and I think it'd be a good idea to remove them and instead recentre the regular gloves to fit the slim model better instead.

bconlon1:
closes #1441


OzPayn:
When you don't have access to the dev halo or glow now the customisation menu is completely empty, would it be worth hiding it entirely for users who don't have any options in there?

OzPayn (approved):
wonderful

PR #1463, opened 2023-05-30
feat: Add debug mcfunctions

Burning-Cactus:
Functions in Minecraft are a convenient way to add developer utilities without exposing them to the average player. This PR adds a couple of functions from Twilight Forest, as well as our own for instantly teleporting to the Aether and suiting up with equipment.


bconlon1 (approved):
lgtm

PR #1482, opened 2023-06-05
feat: added Randomium support

GizmoTheMoonPig:
[Randomium](https://www.curseforge.com/minecraft/mc-mods/randomium-ore) is a small mod that adds an ore that can drop basically any item. They have quite a few different checks for creative only items, but of course no system is perfect and certain Aether blocks can drop from the ore even though they really shouldn't. This PR exists to prevent dungeon blocks (locked, trapped, boss doorway, and treasure doorway), chest mimics, and treasure chests from dropping from the ore. 

I don't know how you guys feel about mod support like this, but I figured since it was just a single tag it would be ok. Feel free to close this if you aren't interested. 

---


bconlon1 (approved):
lgtm

PR #1483, opened 2023-06-05
fix: netherite glove smithing now works with the 1.20 experimental pack

GizmoTheMoonPig:
Mojang changed how the smithing table works in 1.20. This means that the current recipe to turn diamond gloves to netherite ones doesn't work if the 1.20 experimental pack is enabled. This PR adds support for the 1.20 method of smithing while still keeping the old recipe for people who don't use the experimental pack. 

---


bconlon1 (approved):
lgtm

PR #1496, opened 2023-06-09
fix: The slider doesn't emit walking sounds

### LINKED ISSUE #1480: Bug: Slider makes footstep sounds

Entity

### Other Type

_No response_

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

45.0.64

### The Aether Version

beta.4

_No response_

### Client Log

_No response_

_No response_

### Steps to Reproduce

1. Hit the Slider with a Pickaxe in survival mode
2. Switch to spectator
3. Make the slider follow you into the ground, making it touch the floor (don't go too low so it doesn't reset)
4. Make the slider move around the dungeon, and listen closely and hear footstep sounds

Alternative, Survival-only steps:
1. Awaken the slider
2. Get it to touch the floor
3. Listen carefully to it as it glides across the floor while touching the floor

### What You Expect To Happen

There wouldn't be footstep sounds, since the Slider is meant to slide.

### What Actually Happened

The Slider made footstep noises when gliding across the floor

### Additional Details

_No response_

Burning-Cactus:
Closes #1480

### DIFF
--- a/src/main/java/com/aetherteam/aether/entity/monster/dungeon/boss/slider/Slider.java
+++ b/src/main/java/com/aetherteam/aether/entity/monster/dungeon/boss/slider/Slider.java
@@ -522,6 +522,12 @@ public class Slider extends PathfinderMob implements BossMob<Slider>, Enemy, IEn
         return false;
     }
 
+    // The slider should not be making footstep sounds.
+    @Override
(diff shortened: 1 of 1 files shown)


bconlon1 (approved):
lgtm

Issue #1534, opened 2023-06-17
Feature Request: include the RecipeType in ItemUseConvertEvent


Zepalesque:
### What Feature Types Apply to This Request?

System

### Is This an Enhancement to a Preexisting Feature?

- [X] I confirm this request is an enhancement to a feature already in The Aether.

### What Is Your Feature Request?

For the ItemUseConvertEvent to include a field/methods to get the RecipeType of the conversion

### Why Should It Be in the Mod?

So that addons using the event can know what kind of conversion took place

For instance, I am trying to add unique sounds to each type of item use conversion, and having the RecipeType would allow me to do that without needing to guess based on the old and new blockstates.

### Please Read and Confirm the Following

- [X] I have confirmed this feature is either in the original Aether mod, or improves parity between The Aether and Minecraft or other mods.
- [X] I have confirmed this suggestion is unique and has not been suggested already.

PR #1535, opened 2023-06-18
feat(recipe): Added RecipeType to ItemUseConversionEvent

Zepalesque:
This pull request adds the RecipeType to the ItemUseConversionEvent, so that addons can use it in their event handlers.

closes issue #1534 

---


bconlon1 (approved):
lgtm

PR #1539, opened 2023-06-18
fix: ItemUseConvertEvent running when it shouldn't

### LINKED ISSUE #1536: Bug: ItemUseConvertEvent does not function correctly

Block, Item, System

### Other Type

_No response_

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

45.0.64

### The Aether Version

70a6977

_No response_

### Client Log

_No response_

_No response_

### Steps to Reproduce

1: Open Minecraft in the debugger
2: Put a breakpoint in ItemUseConvertEvent
3: Use Ambrosium or a Swet Ball on a block that it does not convert
4: Observe the value of `newBlockState` in the debugger
5: Attempt the same with a Swet ball on something it _can_ work on, like Aether Dirt and observe the `newBlockState` value again

### What You Expect To Happen

The event would not fire the first time, and the second time, it would have the correct `newBlockState` value

### What Actually Happened

For the first one, the event fired regardless of the fact that it is an invalid recipe target. It also used `aether:enchanted_aether_grass` for the `newBlockstate` when using ambrosium and `minecraft:crimson_nylium` when using a swet ball.

For the second test of a Swet Ball on Aether Dirt, it properly fired the event like it should, but the `newBlockState` value showed up as `minecraft:crimson_nylium` for some reason.

### Additional Details

I _would_ fix this in my pull request, but I'm not sure what's wrong whatsoever, the code looks fine to me. It also goes a bit outside my knowledge unfortunately.

bconlon1:
closes #1536

### DIFF
--- a/src/main/java/com/aetherteam/aether/item/materials/behavior/ItemUseConversion.java
+++ b/src/main/java/com/aetherteam/aether/item/materials/behavior/ItemUseConversion.java
@@ -31,13 +31,15 @@ public interface ItemUseConversion<R extends MatchEventRecipe & BlockStateRecipe
         for (R recipe : level.getRecipeManager().getAllRecipesFor(recipeType)) { // Gets the list of recipes existing for a RecipeType.
             if (recipe != null) {
                 BlockState newState = recipe.getResultState(oldBlockState); // Gets the result BlockState and gives it the properties of the old BlockState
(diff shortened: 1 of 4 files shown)


Zepalesque:
hmm, I noticed something odd with the latest commit of this pull request, it seems like regardless of whether or not it actually has a recipe, ambrosium and swet balls are doing the arm swing animation on every block when right-clicked, although without posting the event (as expected, the only odd part is the arm swing)

raptor494 (approved):

PR #1568, opened 2023-07-08
fix: If the moaType is null, use the blue type as fallback for save data

carl-vbn:
I have a world in which I experienced frequent crashes with the following error:
```
The game crashed whilst saving entity nbt
Error: java.lang.NullPointerException: Cannot invoke "com.aetherteam.aether.api.registers.MoaType.toString()" because the return value of "com.aetherteam.aether.entity.passive.Moa.getMoaType()" is null
```

I noticed that in `Moa#addAdditionalSaveData`, the `toString()` method is called on `getMoaType()` without making sure the method returns a non-null value. I also noticed that the return value of `getMoaType()` is treated as potentially null in many places around the Moa class, and that the blue Moa type is often used as a fallback for those cases. I therefore extended this pattern here, and after some testing it seems like the crashes no longer occur. 

---


bconlon1:
Were you running into this issue with the latest beta? No one has run into this issue for awhile, I'm surprised it still exists.

bconlon1 (approved):
lgtm

carl-vbn:
I built the latest code earlier today and played the affected world with it and still had the crash. I am playing with other mods on that world though, but the crash reports only mentions the Aether.

bconlon1:
It could be from an incompatibility of some sort then, the fix seemed like a good precaution either way though so it will be in the next beta.

carl-vbn:
Thank you. It seemed right to me since the code is already doing something similar in various other places in that class.

PR #1569, opened 2023-07-08
improv(structure): Gold Dungeon grass is dependent on surface rule.

### LINKED ISSUE #1549: Feature Request: Make gold dungeons use Surface Rules for their surface/grass blocks

Block, System, World-Gen

### Other Type

_No response_

For the golden dungeons to use the surface rules of the current biome to choose their surface blocks, rather than always using Aether Grass Blocks

So that addons that add new biomes with alternate surface rulea can still spawn the gold dungeon in their biome without it looking too out-of-place (since I’m pretty sure it’s supposed to at least *partially* blend in with the environment, just look a *little* bit suspicious, and if it has an entire different type of grass block then it would be pretty obvious that it’s a generated structure for new players)

Addons that add new biomes with alternate surface rules

bconlon1:
closes #1549


OzPayn (approved):
seems to work for me

PR #1571, opened 2023-07-08
refactor: Manually cancel BossEventProgress after it is posted

### LINKED ISSUE #1570: Bug: Jade's boss bar pushdown does not respect Aether boss bars

Gui

### Other Type

_No response_

### What Type of Bug Is This?

Compatibility

### Forge Version

45.1.0

### The Aether Version

git commit c9f4c5eed9da34e114fab5ce208691a7ea9fab3d

Jade, 10.0.0

### Client Log

_No response_

_No response_

### Steps to Reproduce

1. Run the game with Aether and Jade.
2. Trigger any Aether-specific boss battle.

### What You Expect To Happen

Jade pushes itself down below the boss bar.

### What Actually Happened

Jade does not move from the top of the screen.

### Additional Details

I've tested this in the dev environment using the provided Jade version within `gradle.properties`. Feel free to assign me, I already have a working PR that is undergoing testing right now.

Original report by noctu on Discord.

relevant conversation

Jonathing:
Closes #1570.

Jade's boss bar pushdown works by checking if `CustomizeGuiOverlayEvent.BossEventProgress` is canceled. This means that, technically, we're the ones breaking the rules by disabling the normal boss bar and rending our own on top of it.

Instead of refactoring Aether's boss bar system to use a mixin into `BossHealthOverlay` based on the results of the event (although you can do that if you want if you decide that that is a better solution over this), this PR's solution is to allow the event to play out to all listeners before setting the event as canceled after the event result has been stored in its variable (see `BossHealthOverlay:37`). That way, event listeners can act as if the Aether's boss bar is the normal boss bar (since we already set the increment in our listener) without needing to rely on the event being canceled.

proof that it works

### DIFF
--- a/src/main/java/com/aetherteam/aether/client/event/listeners/GuiListener.java
+++ b/src/main/java/com/aetherteam/aether/client/event/listeners/GuiListener.java
@@ -122,7 +122,8 @@ public class GuiListener {
 		if (BOSS_EVENTS.contains(bossEvent.getId())) {
 			GuiHooks.drawBossHealthBar(event.getPoseStack(), event.getX(), event.getY(), bossEvent);
 			event.setIncrement(event.getIncrement() + 13);
-			event.setCanceled(true);
+			// This event is cancelled in BossHealthOverlayMixin. see it for more info.
+			//event.setCanceled(true);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


bconlon1 (approved):
lgtm

PR #1572, opened 2023-07-08
build: Cleanup jar artifact building and publishing pointers

Jonathing:
This is another Gradle cleanup PR that I'd like to propose. I noticed that with the way your `build.gradle` system is set up, the `build` and `publish` tasks *must* act in tandem with each other. This can also be cause for concern if there is a desync in the source code but there is no check for it in the `publish`, since it could publish the old jar file without checking sources *and* properly reobfuscating the jar file.

To solve this, I've allowed the `jar` and `jarJar` tasks to act independently of each other. In the publishing configuration, I've added `artifact project.tasks.jarJar`, which will allow Gradle to grab the file specified manually by the `jarJar` task. Additionally, I've configured the base `jar` task to have a classifier of "no-embeds". This leaves the normal, non-embedded jar file intact in case you want to play around with it more for debugging purposes without the need to override it with the embedded jar file.

Finally, I've added two lines to force Gradle to ensure that the "jar" and "jarJar" archive tasks are always followed up by their reobfuscation counterpart. This might solve any additional reobfuscation problems with building the project in the future.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -147,6 +147,10 @@ tasks.jarJar.configure {
     archiveClassifier = ""
 }
 
+// ensure jars are reobf'ed before confirming artifact files
+jar.finalizedBy('reobfJar')
+tasks.jarJar.finalizedBy('reobfJarJar')
+
 tasks.named(sourceSets.main.compileJavaTaskName) {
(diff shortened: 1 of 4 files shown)


bconlon1 (approved):
lgtm

PR #1573, opened 2023-07-08
build: Publish an obfuscated sources jar

Jonathing:
Publishing an obfuscated sources jar file solves the need to manually specify to Gradle to only deobfuscate the runtime and use the compile normally. ForgeGradle has the capability to deobfuscate source jar files and it does so rather well. This also allows addon developers to work with other mappings of choice (which may be important for those who still use Yarn, for whatever reason).

This is done using [ModGradle's SourceJar](https://github.com/ModdingX/ModGradle/blob/master/docs/sourcejar.md) Gradle plugin, which does all of the heavy lifting for you (including creating the "sourceJar" task). I've edited the README file to account for this change if you wish to implement it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -155,8 +155,8 @@ publishing {
     publications {
         mavenJava(MavenPublication) {
             artifactId project.archivesBaseName
-            artifact jar
-            artifact sourcesJar
+            artifact project.tasks.jarJar
+            artifact sourceJar
         }
     }
     repositories {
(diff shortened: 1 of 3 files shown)


bconlon1 (approved):
lgtm

bconlon1 (changes requested):
Resolve merge conflicts.

bconlon1 (approved):
lgtm

bconlon1 (changes requested):
un-lgtm

bconlon1 (approved):
re-lgtm

PR #1575, opened 2023-07-09
feat(config): Aether Start

### LINKED ISSUE #1557: Feature: Start In Aether config for modpacks only
- Include disclaimer that other configuration would be necessary for this to not potentially mess with the player's ability to stay in the dimension without a portal.

bconlon1:
closes #1557

### DIFF
--- a/src/main/java/com/aetherteam/aether/capability/player/AetherPlayerCapability.java
+++ b/src/main/java/com/aetherteam/aether/capability/player/AetherPlayerCapability.java
@@ -115,6 +115,7 @@ public class AetherPlayerCapability extends CapabilitySyncing implements AetherP
 	public CompoundTag serializeNBT() {
 		CompoundTag tag = new CompoundTag();
 		tag.putBoolean("CanGetPortal", this.canGetPortal());
+		tag.putBoolean("CanSpawnInAether", this.canSpawnInAether());
 		tag.putFloat("SavedHealth", this.getSavedHealth());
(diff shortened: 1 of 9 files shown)


Jonathing (changes requested):
Unless the user has Configured installed (whose GUI configs allow for default server configs on world creation), it will be very hard for a client, who wants to create a single player world, to have this config boolean magically enabled. It might be prudent to create a built-in datapack that simply triggers spawning the player in the Aether as well.

bconlon1:
@Jonathing 
This config isn't intended for datapack makers and not really average players creating a world, as is said in the config comment. Either way though, this is still able to be set with `defaultconfigs` by anyone.

src/main/java/com/aetherteam/aether/capability/player/AetherPlayerCapability.java
```diff
@@ -46,6 +46,7 @@ public class AetherPlayerCapability extends CapabilitySyncing implements AetherP
 	private static final UUID LIFE_SHARD_HEALTH_ID = UUID.fromString("E11710C8-4247-4CB6-B3B5-729CB34CFC1A");
 
 	private boolean canGetPortal = true;
+	private boolean canSpawnInAether = true;
```
bconlon1, line 49: This value needs to be set to NBT.

Jonathing:
Well, the modpack developers can figure it out. 😃👍

bconlon1:
Meant to say *is intended for modpack players.

raptor494 (approved):

PR #1589, opened 2023-07-14
build: Upgrade to ForgeGradle 6 and switch to Sugarcane mappings

Jonathing:
I tried to get ModGradle's mapping plugin to work on ForgeGradle 5, but I kept getting weird build errors, so I decided to solve the problem by upgrading the entire buildscript anyway.

Buildscript upgrade details:
- Upgraded to Gradle 8.1.1.
- Gradle's Eclipse plugin has been updated and has returned to the buildscript.
- Upgraded to ForgeGradle 6.
- ForgeGradle 6 uses Gradle's foojar toolchain convention to work with java launchers. This is applied in `settings.gradle`.
- MixinGradle is now properly published as a modern Gradle plugin, and can be used in the plugin DSL without additional setup in `settings.gradle`.
- MixinGradle now no longer causes issues when the Eclipse plugin is applied.
- Archive base name is now set through the `base` closure block.
- ModGradle's SourceJar plugin needs to be applied manually after the Java language version is set (see ModdingX/ModGradle#17)
- ForgeGradle has introduced new options within the `minecraft` block. I have added them in with their default values.
- Run configurations can now have their base values properly configured for each.
- Added the `gameTestServer` run configuration (this has been here for a while but did not come included in ForgeGradle 4).
- Tasks can now be configured while casted to their proper type (ex. `Jar`, `JarJar`).

Sugarcane migration details:
- Sugarcane mappings are simply Parchment mappings but with sanitized parameter names (no more `p` prefix in parameter names).
- Sugarcane is automatically built for any Parchment release. The version schema is the same as Parchment's (see `gradle.properties`).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - setup
       - run: |
-          ./gradlew publish
+          gradle publish
 
 commands:
   setup:
@@ -26,7 +26,7 @@ commands:
       - gradle/with_cache:
           cache_key: 'v9'
           steps:
-            - run: ./gradlew build --parallel --build-cache --console=plain
+            - run: gradle build --parallel --build-cache --console=plain
       - store_artifacts:
           path: ~/project/build/libs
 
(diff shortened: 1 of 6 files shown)


bconlon1 (approved):
lgtm

Issue #750, opened 2022-08-17
Chore: Mixin Cleanup


Jonathing:
# Mixins

are bad!

## Anyways

Mixins are actually great (I've been trolling you the whole time). They provide a stable API to modifying bytecode that we otherwise wouldn't be able to. Of course, using Mixins still require some knowledge of how bytecode works to properly use them, but nonetheless Mixins are so easy and useful that they remind me of the ~~glory~~ days of jarmodding.

However using Mixins (or really any form of coremodding) can be really dangerous, because they pose the threat of causing behaviors that would otherwise not be accounted for. The reason for this is because Forge's entire existence relies on mod compatibility through the use of the Forge API, unlike stripped down modloaders like Fabric which rely heavily on the usage of coremodding. It's because of this that mixins should be written as cleanly, efficiently, and effectively as possible, while also keeping their numbers as low as possible so as to not add behaviors that could otherwise break the fuck out of the game.

## Disclaimer

I'm an asshole. I apologize if I come off as rude in this issue, but I want to be as thorough as possible with my critiques of these mixins. Remember that faulty coremodding can effect everybody, not just the mod at hand.

### General Points

There are a couple of general points that apply to either more than one mixin, or are general enough that they should always be followed as a rule of thumb.

- If you need to target an `INVOKE` instruction and inject code **after** that instruction, *do not* use `@At(value = "INVOKE", shift = At.Shift.AFTER)`. Instead, use `@At(value = "INVOKE_ASSIGN"`.
  - Actually to add on to that, I do not recommend using `shift` altogether. Always attempt to target a specific instruction unless you have ground to shift the injection point otherwise.
- Although this doesn't cause any issues at the current moment, I would *not* recommend using only the method name in the `method` parameter of `@Inject` or similar annotations. Always include the full method description as well. IntelliJ has a built-in Bytecode Viewer so you can easily copy the description from the viewer, but you can also copy the entire Mixin target reference in `Right Click -> Copy / Paste Special -> Mixin Target Reference`. Unlike the method description, the method owner (the class name) is not required.

### :heavy_check_mark: ItemStackMixin

This seems like kind of a redundant mixin. You can easily modify an itemstack's default attributes by overriding the `getDefaultInstance()` method in the itemclass, and providing your own default itemstack.

- Additionally, I'm getting a compiler warning saying that it can't find the srg name for a forge method. In the `@At()` annotation, be sure to mark any non-Minecraft methods as `remap = false`.

[...]

PR #1604, opened 2023-08-05
improv: Valkyrie tool reach base-level functionality and a few stylistic cleanups

Jonathing:
Successor to #750. I left a lot of notes on the actual commits themselves, but basically this is my final note on the mixin cleanup. The most important of which is that `LivingEntityMixin` now injects into `getAttributeValue()` to check if it should modify the value for reach, depending on if the player is attempting to access the attribute in question and is holding a valkyrie tool. It uses the same logic as before, but now it is more concise and affects the game at the base-level, so there is no need to subscribe to multiple events which would have run this calculation on its own.

Here is the full diff of that specific change, without the additional formatting refactors I did: https://github.com/The-Aether-Team/The-Aether/compare/b1ea57141fbbd4a0cd3af70cfb1f2b968d27e94a...b06a3f5651d696638880e5a2a59bc19412877ef7

Please read each of the commit notes for full notes and context on each change I made here. I made sure to take great care in refactoring and repurposing these mixins.


bconlon1 (approved):
lgtm

PR #1609, opened 2023-08-08
feat: CraftTweaker compatibility for fuel editing

### LINKED ISSUE #288: Feature: CraftTweaker compatibility.
- [x] Reimplement CraftTweaker in gradle to the 1.19 branch.
- [x] Removing and adding custom fuels.

bconlon1:
closes #288

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -116,9 +116,6 @@ dependencies {
 
     compileOnly fg.deobf("top.theillusivec4.caelus:caelus-forge:${project.caelus_version}")
 
-    compileOnly "mezz.jei:jei-${project.mc_version}-common:${project.jei_version}"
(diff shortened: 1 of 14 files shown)


bconlon1:
Documentation: https://github.com/CraftTweaker/CraftTweaker-Documentation/pull/549

jaredlll08:
In general we (CraftTweaker) generally ask mods to not add new globals, as it will cause issues if 2 mods try and use the `altar` global.

My recommendation would be to instead make the methods static and have a static call like so:
```zenscript
mods.aether.AltarFuelManager.addFuel(<item:minecraft:dirt>, 200);
```

bconlon1:
Updated methods to be static and removed globals accordingly.

OzPayn (approved):
should be good to merge

PR #1613, opened 2023-08-09
improv: Better system for culling Aerogel block faces

### LINKED ISSUE #1608: Bug: Spam-placing Aerogel Stairs causes block renders to stop updating
Attempting to close the game also fails.

bconlon1:
closes #1608


OzPayn (approved):
couldnt find anything broken so I think this is good

PR #1617, opened 2023-08-11
feat: Remedy Effect

### LINKED ISSUE #1616: Feature Request: Make Remedy an effect
- Maybe also make it negate inebriation while its active. maybe also tie the duration of the overlay to the duration of the effect.

bconlon1:
closes #1616 

- Fixes Inebriation having particles.


OzPayn (approved):
I dig it

PR #1657, opened 2023-08-22
refactor: Tool debuff rebalancing

### LINKED ISSUE #1655: Bug: Tool debuff is oftentimes too aggressive
Some tool interactions end up slower than mining by hand. Couple players have suggested reworking the tool debuff to be less aggressive. Might be worth exploring rebalance/redesigns potentially.

bconlon1:
closes #1655


reetamb (approved):

OzPayn (approved):
feels correct to me

PR #1678, opened 2023-09-07
feat: Initial update to 1.20.1

CDAGaming:
This is an initial diffset for the 1.20/1.20.1 update, from both myself and @Camotoy

This port is functional save for a few things:

- There is an issue where, in select cases, multiple "tips" may render in the same position
- There are several instances of `RenderSystem` calls that may be migratable to `GuiGraphics` equivalents
- The Aether Title Screen Splash Text is currently stubbed and not-present
  - I have ported the logic for the Vanilla-Left Splash Text, though ran out of time to do the same fix for the Aether Title Screen

@Camotoy was responsible for the initial diffset, while I took hold of resolving the issues the initial diffset had, in addition to migrating the datagen-related logic and release metadata (Using Cumulus and Nitrogen Repos as a reference).

This diffset has been combined into one commit, though the original set of commits can be viewed on the non-combined branch (Git has messed up the commit times from the rebase, but work on this originally started around 1-2 months back, beginning with Camotoy's original commits)

---

I (And @Camotoy) agree to the Contributor License Agreement (CLA).


GizmoTheMoonPig:
Please make sure you guys properly replace your block materials instead of just outright deleting them. Most materials now use a series of block properties to replicate how they used to act, I have them all listed [here](https://gist.github.com/GizmoTheMoonPig/77a90a48e0aeecd15b4c524e1c7f0a4a). Most mod ports I've come across don't actually do this, and if you don't your blocks may not act as expected. (examples: logs and leaves catching on fire, leaves being destroyed by pistons, various blocks changing the instrument on a noteblock)

Camotoy:
I believe I did use that guide, but maybe not to its fullest extent, so thank you regardless.

I agree to the Contributor License Agreement.

CDAGaming:
@GizmoTheMoonPig Thanks for the tip! Currently rebasing towards the conflict fixes, then I'll sift through AetherBlocks again to apply this change (Thanks for the list, that helps a bunch for this!)

CDAGaming:
Done! AetherBlocks should now match up closely to the list provided.
@bconlon1 will likely want to take a second look into this class, since there's definitely some cleanup that could be done there (And for how they would like to handle the more complex replacements mentioned on the list)

PR #1692, opened 2023-09-10
fix: PVP config works correctly again

### LINKED ISSUE #1688: Bug: Invisibility Cloak for PVP config doesn't seem to work
In both 1.19.2 and 1.19.4 I can't seem to get the invisibility effect to stop, weirdly my Halo becomes visible but my body does not. Mobs react properly so the invisibility effect for PvE is dropping but the visual aspect isn't working correctly

bconlon1:
closes #1688


OzPayn:
no it doesnt 😂

OzPayn (approved):
Yes it does!

PR #1694, opened 2023-09-12
refactor(event, entity): Made EggLayEvent use an ItemStack instead of an Item

Zepalesque:
This pull request makes it so that the `EggLayEvent` stores an `ItemStack` instead of an `Item`, so that more information is given AND can be changed. For instance, I want to be able to store the `MoaType` in an NBT tag for the dropped item so that it can be used without needing to add a new item for each `MoaType` (which considering that other mods can add new `MoaTypes`, probably wouldn't be universally possible anyway)

---


bconlon1 (changes requested):
Documentation needs to be updated too.

Zepalesque:
oh yeah, good point, fixed

bconlon1 (approved):
lgtm

PR #1703, opened 2023-09-14
Update copper_lantern_conversion.json

MehVahdJukaar:
just noticed this while scrolling through this repo to see what the dimension ID is for mod compat. Anyways moved mod ids as the block itself was moved to supplementaries squared


bconlon1 (approved):
lgtm

PR #1704, opened 2023-09-15
refactor: Moa UV map expanded

OzPayn:
Changed the Moa/cockatrice UV map to separate duplicated elements and allow asymmetrical designs


bconlon1 (approved):
lgtm

Issue #1711, opened 2023-09-17
Bug: Moa Donator Screen text is horribly positioned


GizmoTheMoonPig:
### What Feature Types Apply to This Bug?

Entity, Gui

### What Type of Bug Is This?

Annoyance

### Forge Version

47.1.3

### The Aether Version

1.20.1-1.0.0-beta-1

### Steps to Reproduce

1. open inventory
2. open aether customization screen (the one where you can equip gloves and rings and stuff)
3. hit the moa skins button
4. Hover over any component in the moa skin screen
5. observe that all text is rendered in the bottom right corner of the screen instead of in the tooltip bubble

### What You Expect To Happen

The text would appear where my mouse was hovering and would be decently legible

### What Actually Happened

Text always renders in the bottom right corner, no matter what gui size you have selected. Text also renders a darker form of the text over it slightly offset, looks like someone forgot to turn off shadows on text :P

### Additional Details

Attached is an image of this in action. ![Moa Screen](https://github.com/The-Aether-Team/The-Aether/assets/67468252/d895a714-1a6b-4b0b-9096-227e5a2152af)

Note that in order to disable the shadow rendering on text you will have to add a false boolean to each drawString call in guigraphics. I don't know why Mojang didn't make that the default, but it is what it is. 

Also, don't think it's worth another bug report, but hovering over the moa skins and customization buttons renders the tooltip in the bottom right corner of the mouse instead of the top right. Minecraft uses the top right everywhere in their code so I figure it might be good to fix that to make it consistent. Just a small nitpick though. 
(also happens on the moa skin refresh button and the undo and save buttons in the customization screen)

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

PR #1714, opened 2023-09-17
fix(gui): Moa Skin Tooltips render properly again

GizmoTheMoonPig:
Fixes Moa Skin tooltips rendering all the way in the right corner. This was done by switching the Moa rendering PoseStack to use the GuiGraphics one instead of the RenderSystem one. Because of this, this PR also reduces the amount of copied vanilla code. The actual vanilla methods are used now, which means it should hopefully not break again (fingers crossed)

closes issue #1711 

---


bconlon1 (approved):
lgtm

Issue #1715, opened 2023-09-17
Bug: invis cape effect not woking when someone joins your world on 1.20.1


Alexmasonseals:
### What Feature Types Apply to This Bug?

Entity, Item

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

47.1.47

### The Aether Version

beta 6 1.20.1

### Is This Bug a Conflict With Another Mod?

no

### Steps to Reproduce

equip an invisibility cape
get someone to join your world

### What You Expect To Happen

the 2nd player not being able to see you due to the invis cape

### What Actually Happened

2nd player being able to see you until you disable and re-enable the invis cape

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

PR #1723, opened 2023-09-17
improv: Invisibility no longer desyncs on server

bconlon1:
#1715


OzPayn (approved):
looks good at a glance, hopefully that fixes it

PR #1727, opened 2023-09-18
fix: Proper moa UVs in legacy packs

bconlon1:
#1726


OzPayn (approved):
ate my lunch now I can approve this message

even though I technically contributed

but also this was correcting my own mistake so idk

approved

Issue #1722, opened 2023-09-17
Bug: Vanilla and Aether music managers interact oddly


KingPhygieBoo:
### What Feature Types Apply to This Bug?

Audio, System

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

47.1.0

### The Aether Version

1.20.1-1.0.0-beta.1

### Steps to Reproduce

Switch gamemodes from creative to survival. this causes the ``currentMusic`` value in AetherMusicManager to be nulled out without stopping

### Side effect that can be reproduced in-game
- Start in creative, and allow music to play
- Switch to survival. If the vanilla music manager's timer hits 0, it can play music on top of what the Aether's played (this does not work vise-versa)

### What You Expect To Happen

``AetherMusicManager.currentMusic`` should either continue to have a value assigned while playing, or the sound playing should be stopped when the value is nulled.

### What Actually Happened

``AetherMusicManager.currentMusic`` is nulled, but the song continues to play, and cannot be referenced directly anymore. Music can overlap due to this

### Additional Details

Video showing overlap occuring
https://github.com/The-Aether-Team/The-Aether/assets/17862210/61ab9ae3-8281-4042-a475-fbcec8ab5275

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

bconlon1:
Note for how vanilla handles this: switching from creative to survival doesn't stop the currently playing music, so I will have to think of a way to fix this that isn't just stopping the music when the value isn't null.

PR #1728, opened 2023-09-18
fix: Only set music to null after no longer active

bconlon1:
#1722


OzPayn (approved):
seems to be working, approved

PR #1735, opened 2023-09-19
feat: dart shooter balance experiment

OzPayn:
reduced firing speed
slightly reduced projectile speed
reduced damage
reduced power enchantment modifier

### DIFF
--- a/src/main/java/com/aetherteam/aether/entity/projectile/dart/EnchantedDart.java
+++ b/src/main/java/com/aetherteam/aether/entity/projectile/dart/EnchantedDart.java
@@ -8,11 +8,11 @@ import net.minecraft.world.level.Level;
 public class EnchantedDart extends AbstractDart {
     public EnchantedDart(EntityType<? extends EnchantedDart> type, Level level) {
         super(type, level, AetherItems.ENCHANTED_DART);
-        this.setBaseDamage(2.5);
+        this.setBaseDamage(1.5);
     }
 
     public EnchantedDart(Level level) {
         super(AetherEntityTypes.ENCHANTED_DART.get(), level, AetherItems.ENCHANTED_DART);
-        this.setBaseDamage(2.5);
+        this.setBaseDamage(1.5);
     }
 }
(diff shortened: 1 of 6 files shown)


src/main/java/com/aetherteam/aether/entity/projectile/dart/EnchantedDart.java
Sto3IV, line 1: from 6.0 to 4.0 will be more fine

src/main/java/com/aetherteam/aether/entity/projectile/dart/GoldenDart.java
Sto3IV, line 1: no

src/main/java/com/aetherteam/aether/entity/projectile/dart/PoisonDart.java
Sto3IV, line 1: no. current numbers if fine

src/main/java/com/aetherteam/aether/item/combat/DartShooterItem.java
Sto3IV, line 1: depends on Minecraft difficulty will be better.

Sto3IV (changes requested):
easy difficulty = ez mode
hard difficulty = please no. no darts.... i'm falling. ded

OzPayn:
> easy difficulty = ez mode hard difficulty = please no. no darts.... i'm falling. ded

Difficulty settings don't generally affect player functionality in vanilla, I do not think that would be a good idea.

The damage numbers Ive chosen are based on the amount of hits required to kill some of our tankier mobs like Valkyries. If the value is too high the dart shooters very very quickly outcompete even the highest enchanted swords. The values I've proposed here put the dart shooters just below swords in terms of damage output, but they also still retain the very impactful strength of being zero-gravity projectiles. So I think there would be an argument for perhaps debuffing them further. But I'd like people to actually test these values first and see how they feel rather than just commenting on the numbers themselves.

bconlon1 (approved):
lgtm

PR #1769, opened 2023-09-24
Update Moa saddle layer

OzPayn:
Changed the saddle layer to be expanded across the whole model, results in better visuals for supporter moa skins


bconlon1 (approved):
lgtm

PR #1783, opened 2023-10-01
improv(gui): Change Pro Tip display if "Tips" mod is installed

GizmoTheMoonPig:
If you install Aether and [Tips](https://www.curseforge.com/minecraft/mc-mods/tips) together, both mods will display their tips, often times overlapping. This PR makes it so Aether's Pro Tips render in the Tips style when the mod is installed. Since tips can be added via resource packs, I made it so tips are datagenned along with the lang entries for them, meaning if any more pro tips were to be added in the future they would automatically be added as tips. 

I probably need a little feedback if this is going to be merged. I currently have the tip title display as "Pro Tip:", as shown below. This system gives us complete control over how the title and tip itself render. The color and font style (bold, underline, etc.) can all be changed. If there's some other way you'd like to see the title and/or tip rendered, please let me know. Tips just displays the title as "Tip", if this is preferred, I can easily change it to do that instead.
Aether Tip

Feel free to deny this if it's not wanted as well. It was a super easy PR to make so no harm done if it's not accepted. I figured this would be a nice compat feature, but I understand if it changes too much for your liking.

---


bconlon1 (approved):
lgtm

Issue #1785, opened 2023-10-01
Feature Request: Make each Boss use their own Boss Bar Texture instead of having them all use one 


Builderdog841:
### What Feature Types Apply to This Request?

Art

### Is This an Enhancement to a Preexisting Feature?

- [X] I confirm this request is an enhancement to a feature already in The Aether.

### What Is Your Feature Request?

To have each Boss Bar induvidually customizable using Texture Packs or other Mods. These could all be on the same sprite sheet to reduce file size

### Why Should It Be in the Mod?

It would be helpfull for people like me that would like to redo the Boss Bars and probably also going to be useful later on for Aether Genesis.

### Please Read and Confirm the Following

- [X] I have confirmed this feature is either in the original Aether mod, or improves parity between The Aether and Minecraft or other mods.
- [X] I have confirmed this suggestion is unique and has not been suggested already.

PR #1790, opened 2023-10-04
refactor!: Separate boss bar files

bconlon1:
#1785


OzPayn (approved):
yep those work (tested with editing the health bars to ensure they're correct)

Issue #1781, opened 2023-09-29
Bug: Valkyrie tool bonus range stops working


Cinex20:
### What Feature Types Apply to This Bug?

Item

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

47.1.44

### The Aether Version

1.0.0-beta.1.3-neoforge

### Steps to Reproduce

Happens consistently.
1. Put any item in the off-hand slot.
2. Put any valkyrie tool in the main hand.
3. Aim at any block beyond standard range.
4. Press the right mouse button.

### What You Expect To Happen

Absolutely nothing.

### What Actually Happened

The bonus range stopped working for blocks, but remained unchanged for entities.

### Additional Details

Equipping another valkyrie tool in the off-hand slot will temporarily restore the bonus block reach.

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

bconlon1:
Appears to have been caused by https://github.com/The-Aether-Team/The-Aether/pull/1604.

PR #1791, opened 2023-10-04
improv: Fix extended reach

bconlon1:
#1781


OzPayn (approved):
seems to have fixed the issue for me. All seems to be working correctly

PR #1809, opened 2023-10-13
fix: crash with OptiFine

KaiAF:
In the BossHealthOverlayMixin class, it calls to modify a variable that has `CustomizeGuiOverlayEvent.BossEventProgress`. In OptiFine, it uses it's own reflector classes to make it compatible with Forge. This is what is causing the crashes. What I have done is create two new classes, one is to apply the correct mixin based on if OptiFine is found. The other, is a fixed OptiFine mixin class.

The index has been changed to 9 instead of 7. And since in OptiFine the reflector method is an Object, it now calls for and returns an Object.

image

I have tested with and without OptiFine, and it seems to work just fine.

---


bconlon1 (approved):
lgtm

PR #1824, opened 2023-10-21
fix: remove armor when swapping gloves in smithing screen

### LINKED ISSUE #1821: Bug: Armor remains rendered in smithing table after being switched for gloves

Gui

### Other Type

_No response_

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

47.2.1

### The Aether Version

1.0.0-beta.1.3

_No response_

### Client Log

_No response_

_No response_

### Steps to Reproduce

1. Fill a smithing table's template and material slots with any armor trim template and any armor trim material
2. Add any trimmable piece of armor other than gloves into the middle slot
3. Replace the armor piece inside the middle slot with any trimmable gloves

### What You Expect To Happen

The preview of the original armor piece should disappear after the armor item is switched for the gloves

### What Actually Happened

The armor stand in the GUI displays previews of both the gloves and the original armor piece.

### Additional Details

After switching either the trim template or the material, only the gloves' preview changes, the original item's preview doesn't disappear but also doesn't change.

bconlon1:
closes #1821


OzPayn (approved):
yep that fixed it

Issue #1834, opened 2023-11-02
Bug: Plunderer's Remorse advancement is only granted from aether dirt or grass


kodadacod:
### What Type of Bug Is This?

Annoyance

### Forge Version

47.1.0

### The Aether Version

1.20.1 beta 1.4

### Steps to Reproduce

Everytime
1. Till normal dirt/grass with a valkyrie hoe
2. Notice how the advancement wasn't granted
3. Using the same hoe, till aether dirt/grass
4. You now have the advancement

### What You Expect To Happen

You would get the advancement when tilling any type of dirt or grass

### What Actually Happened

You are only granted the advancement when tilling aether dirt or grass

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

PR #1839, opened 2023-11-06
Fix: Plunderer's Remourse Now Triggers On MC Soil

YvanCywan:
Plunderer's Remourse triggers only on tilling of aether soil, added MC vanilla blocks to relevant JSON and Immutable Hashmap.

I've tested this on my local, and is working as intended.

closes issue #1834 
---


bconlon1 (approved):
lgtm

PR #1840, opened 2023-11-06
fix(gui): more fixes to silver heart rendering

GizmoTheMoonPig:
Fixes a couple more issues regarding silver heart rendering. 
- Silver hearts will now properly shake if any are showing and your health is at or below 2 hearts.
  - Do note that if your max health goes into the negatives the shaking will still be incorrect. I can't do anything about this without rewriting the system.
- Silver hearts will no longer render too far to the left if your base max health is put in the negatives. This was very noticeable with [VoidScape's](https://www.curseforge.com/minecraft/mc-mods/voidscape) infusion effect in the void, which shrinks your max health the longer you're in the void. 

---


bconlon1 (approved):
lgtm

Issue #1841, opened 2023-11-06
Bug: Event Hook causing low TPS on Multiplayer Server MC 1.19.2


xkobalx:
### What Type of Bug Is This?

Performance

### Forge Version

43.2.21

### The Aether Version

1.19.2-1.0.0-beta.1.2-forge

### Steps to Reproduce

Only effecting some players.
On player log in.
![image](https://github.com/The-Aether-Team/The-Aether/assets/10780881/a8fdd013-d2f8-4010-b5b8-191f1c2abd98)
This event appears on the profiler and there are constant tps issues.
This event is triggering no matter what dimension the player is in or where they are in those dimensions.
The issue will sometimes disappear for a minute or two before resuming.
The image shows 10% but it has gone as high as 25%.
Effected players log out.
The issue stops immediately.

### What You Expect To Happen

The event should not be constant and should not be using so much of the servers resources.

### What Actually Happened

Event triggering when certain players log in causing TPS issues.

### Additional Details

Issue persists on the effected players between reboots and logins.

Other players have been to the Aether, to the same areas and have completed the same content but are not effected.

Dedicated Server
Server is running Java 17.0.8.1
AMD Ryzen 9 5950x
16GB of 64GB RAM allocated to MC
Running on NVME RAID.

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

Floofle:
Can additionally confirm this, with Forge 43.3.0 and the same Aether version/ no optifine/custom modpack. This issue appears with the config option "Disables falling into the Overworld" set to TRUE and before any player on the server has even entered the aether. 

See(as one example); https://spark.lucko.me/IJLpxSULDc

EDIT: notably, seems to happen to a player that has placed the portal and then not gone in. will update if that fixes the issue.

bconlon1:
I need to rearrange the checks starting here: https://github.com/The-Aether-Team/The-Aether/blob/cc76cf7ebda0f55356623e7a984dc8bf02a45b5f/src/main/java/com/aetherteam/aether/event/hooks/DimensionHooks.java#L174

[...]

PR #1851, opened 2023-11-29
perf: Reorder hook for falling out of the Aether to improve performance

bconlon1:
#1841


OzPayn (approved):
coolio lets get this tested on ANTS

Issue #1843, opened 2023-11-09
Bug: Graves get deleted when fighting Valkyrie Queen


Nites-Path:
### What Feature Types Apply to This Bug?

Entity, Item, Other (Please Describe)

### Other Type

Compatibility?

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

beta.1.4-neoforge

### The Aether Version

aether-1.20.1-1.0.0

### Is This Bug a Conflict With Another Mod?

Simply Graves (also other grave mods and datapacks)

### Steps to Reproduce

Death by Valkyrie Queen while graves are enabled.

### What You Expect To Happen

When the player is dead all attacks should stop destroying blocks preserving the graves.

### What Actually Happened

When she swings the death blow, as it kills the player it also deletes the grave made to store items on death.

### Additional Details

Just my 2 cents, however if when she does this attack it actually breaks the blocks instead of destroying them would solve the issue since the grave will still break, but the items would then be spilled out instead of just deleted. This would still slightly punish players with graves as it would act as a "vanilla" death, but still allow players to get back and get their stuff.

Also a note, I am running on the Direwolf20 1.20 official modpack and reporting the issue also with them.

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

bconlon1:
Technically, this should be solvable on the pack's end by adding graves to the `#aether:valkyrie_queen_unbreakable` tag. However, I should still probably try to see about preventing the Valkyrie Queen from breaking blocks after player death. I should also probably try to add some kind of default check to see if a block is normally unbreakable according to its strength properties, for both the Valkyrie Queen and other bosses.

The Valkyrie Queen's breaking behavior *is* meant to cause broken blocks to also drop items accordingly though. From what I can tell from digging into Simply Graves' code, I can't find behavior for item dropping when a grave is broken, so this might be why this issue is happening, this behavior makes sense though if the grave block is meant to be unbreakable.

PR #1852, opened 2023-11-29
refactor: Valkyrie Queen and Slider check for block strength before trying to break

bconlon1:
#1843


OzPayn (approved):
seems to have fixed it

Issue #1794, opened 2023-10-07
Bug: Causes Cataclysm's Boss Bar Text to be Abnormal


Dr-WeiAL:
### What Feature Types Apply to This Bug?

System

### What Type of Bug Is This?

Annoyance

### Forge Version

47.1.3

### The Aether Version

1.20.1-1.0.0-beta.1.3

### Is This Bug a Conflict With Another Mod?

[L_Enders_Cataclysm-1.39 -1.20.1](https://www.curseforge.com/minecraft/mc-mods/l_ender-s-cataclysm/files/4778613)

### Client Log

[latest.log](https://github.com/The-Aether-Team/The-Aether/files/12837862/latest.log)

### Steps to Reproduce

Just summon any Cataclysm boss.

### What You Expect To Happen

There should be no textual anomalies

### What Actually Happened

The text is overlapping, and the boss bars are probably overlapping too
![image](https://github.com/The-Aether-Team/The-Aether/assets/47446553/043de860-ac02-4b3a-8898-3c129e1b12a3)

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

PR #1857, opened 2023-11-30
fix: Don't affect boss health event cancellation for non-Aether bosses

bconlon1:
#1794


Drullkus (approved):
A well-done fix, LGTM 👍

PR #1863, opened 2023-12-01
improv: Functional Patreon System again.

bconlon1:
- [x] Update Nitrogen and Aether to support reading new data.
- [x] Update and convert data.
- [x] Create a GitHub Actions workflow to keep data updated.

#1849


OzPayn (approved):
mos

Issue #1873, opened 2023-12-05
Feature Request: Give "Pro Tips" more appropriate formatting if Tips mod is installed


Foozey:
### Is This an Enhancement to a Preexisting Feature?

- [X] I confirm this request is an enhancement to a feature already in The Aether.

### What Is Your Feature Request?

Currently if the Tips mod is installed, Aether's "Pro Tips" are moved to the tips pool, that's great! But they don't fit in very well due to the formatting differences. For example, the tips mod structures tips like this:

> **Tips**
> This is a tip.

The Aether's tips are formatted like this:

> **Pro Tips:**
> This is a tip.

It would be cleaner to have either of the following approaches:
1. Rename Pro Tips title to "Tips" if Tips mod is installed. (This is really the ideal solution, as both tip types will merge seamlessly)
2. Remove colon from title and possibly rename for user clarification, for example "Aether Tips" instead of "Pro Tips:".

### Why Should It Be in the Mod?

Simple change but cleaner result for the user experience.

### Does This Feature Improve Compatibility With Another Mod?

Yes, Tips 1.20.1-12.0.4

### Please Read and Confirm the Following

- [X] I have confirmed this feature is either in the original Aether mod, or improves parity between The Aether and Minecraft or other mods.
- [X] I have confirmed this suggestion is unique and has not been suggested already.

Issue #1872, opened 2023-12-05
Bug: If Tips mod is installed, "Pro Tips" will still be added to the pool even if they are disabled in the config


Foozey:
### What Feature Types Apply to This Bug?

Gui

### What Type of Bug Is This?

Annoyance

### Forge Version

47.2.16

### The Aether Version

1.20.1-1.0.0-beta.2-neoforge

### Is This Bug a Conflict With Another Mod?

No

### Steps to Reproduce

1. Install Tips and The Aether
2. Aether tips will be added to tips pool, but what if you don't want them added?
3. Disable Aether tips in the config but discover they are still shown in the tips pool.

### What You Expect To Happen

Remove Aether tips from the tips pool if they are disabled in the config.

### What Actually Happened

Aether tips are still shown in the tips pool regardless of config setting.

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

PR #1875, opened 2023-12-06
feat: Improve Tips compatibility

bconlon1:
Tips support is now handled using a resource pack that must manually be applied from the Resource Packs screen.

#1872 
#1873


OzPayn (approved):
that appears to be working now! good job

PR #1906, opened 2023-12-28
fix: optimize plant recoloring logic

embeddedt:
This PR optimizes the plant recoloring logic Aether uses. In particular, it is pointless to pass a `ColorResolver` returning a constant value to `level.getBlockTint`. The latter method implements biome blending, but the results are determined based off the given resolver. For a constant color returned everywhere, the blending will change nothing and just adds unnecessary overhead. Instead, the desired color can just be returned as-is.


bconlon1 (approved):
lgtm

PR #1907, opened 2023-12-28
fix: avoid doing expensive biome lookup unless required

embeddedt:
Retrieving biomes from worlds is a relatively expensive task, especially when it's done for every entity in the world. In this case, checking that the entity is low enough in the world is astronomically cheaper, and should be done first. We defer the biome check as long as possible.


bconlon1 (approved):
lgtm

bconlon1:
Since you've contributed, could you provide your UUID?

embeddedt:
My UUID is `6b9f0883-3611-4a57-adb4-2da330283887`

Issue #1918, opened 2023-12-30
Bug: Skyroot Buckets Unusable as Furnace Fuel


PraseodymiumSpike:
### What Feature Types Apply to This Bug?

Item

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

47.1.0

### The Aether Version

1.20.1-1.0.0-neoforge

### Steps to Reproduce

1. Get a Furnace, Blast Furnace, or Smoker. It is acceptable to use Creative Mode for this.
2. Get a Skyroot Bucket. Once again, it is acceptable to use Creative Mode.
3. Attempt to place the Skyroot Bucket in the fuel slot of the Furnace, Blast Furnace, or Smoker.

### What You Expect To Happen

I expected it to go into the fuel slot, from which it may Smelt Items. This is expected based on the fact that other Skyroot things like Skyroot Pickaxes may be used for this purpose, and vanilla Bowls may also be used for this purpose.

### What Actually Happened

The Skyroot Bucket did not enter the fuel slot.

### Additional Details

I don't think the modpack has an issue tracker. There is no GitHub link on https://modrinth.com/modpack/aether-official-server.

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

YvanCywan:
Hey @bconlon1, hope you haven't started this yet, I have a fix for this inbound now.

PR #1919, opened 2023-12-30
Fix(SkyrootBucketItem): Skyroot Buckets Now Burn In Furnances

YvanCywan:
Overrides the getBurnTime method from Item abstact class. Enables the use of this as a fuel.

I think this class could do with a refactor though, it is very heavily nested.

closes issue #1918 


bconlon1 (approved):
lgtm

bconlon1:
Since you have contributed, can you provide your UUID?

PR #1922, opened 2023-12-30
chore: move data data to another class (AetherData)

AlphaMode:
Moves data gen related stuff to a new class called AetherData.
This is just mostly to assist in development with fabric to avoid merge conflicts upstream


bconlon1 (approved):
lgtm

Issue #1920, opened 2023-12-30
Bug: Parachutes get stuck on aerclouds


Zepalesque:
### What Feature Types Apply to This Bug?

Block, Item

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

47.1.70

### The Aether Version

1.0.0

### Steps to Reproduce

1: Grab any type of parachute
2: Descend onto an aercloud
3: Observe how the parachute does not break

### What You Expect To Happen

The parachute would break like usual when landing

### What Actually Happened

The parachute did not break, and instead slid around on the cloud

### Additional Details

video of the bug:

https://github.com/The-Aether-Team/The-Aether/assets/60141811/5b48922d-bc72-48cd-9797-9cce8fc8ce2e

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

PR #1946, opened 2024-01-04
fix: Parachutes now detect collision with more blocks.

bconlon1:
#1920


OzPayn (approved):
looks like that fixed the issue

PR #1947, opened 2024-01-04
Fix generic Curios icons being used rather than the charm/ring ones

Mike-U5:
Hey guys! I noticed a minor quirk in the `Use default Curios' menu config` option. The slots registered in the datapack do not specify what icons they use, so they default to the generic curios icons which makes rings and charms indistinguishable.

bad

Curious has icons for ring/charm slots built-in, so it would be a shame not to use them! Here is how it'll look with this datapack change:

good

### DIFF
--- a/src/main/resources/packs/curios_override/data/curios/curios/slots/charm.json
+++ b/src/main/resources/packs/curios_override/data/curios/curios/slots/charm.json
@@ -1,3 +1,4 @@
 {
-  "size": 2
+  "size": 2,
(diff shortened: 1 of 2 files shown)


bconlon1 (approved):
lgtm

bconlon1:
As a contributor, can you provide your MC UUID?

Mike-U5:
> As a contributor, can you provide your MC UUID?

Oh sure it's  
`615ed918-3eaf-4f73-a3c8-631b4d7da0a1`

Issue #1949, opened 2024-01-05
Optimization: optimizing handling of block related recipes to reduce overhead on block updates


shBLOCK:
Currently, `RecipeHooks.checkExistenceBanned()` and `RecipeHooks.sendIcestoneFreezableUpdateEvent` are called for **every** `NeighborNotifyEvent`.
This causes lots of overhead on block updates. (the biome checking of `checkExistenceBanned` is especially bad, costing more than one ms every tick when there are lots of block updates happening)

The call site:
https://github.com/The-Aether-Team/The-Aether/blob/7d8ed40c63ec400074f1e483bbee62c566d20d24/src/main/java/com/aetherteam/aether/event/listeners/RecipeListener.java#L45-L51

I'm not sure if this is intentional or not (if there are block recipes for the dimensions other than the aether), but if not, I think it'd be pretty easy to just add an early-exit condition on the dimension before checking the actual recipes and it'd get rid of this overhead (unless someone decides to build their whole base in the aether).

bconlon1:
I am a bit confused as to what biome checking youre referring to

shBLOCK:
> I am a bit confused as to what biome checking you're referring to

This bit. (`level.getBiome()` is the most time-consuming part)
https://github.com/The-Aether-Team/The-Aether/blob/7d8ed40c63ec400074f1e483bbee62c566d20d24/src/main/java/com/aetherteam/aether/recipe/recipes/block/AbstractBiomeParameterRecipe.java#L40-L48

PR #1950, opened 2024-01-05
Optimize block recipe checking to reduce overhead on block updates

shBLOCK:
closes issue #1949
This would make the **block recipes no longer work in dimensions other than the aether**, check the issue for more details.

### DIFF
--- a/src/main/java/com/aetherteam/aether/event/listeners/RecipeListener.java
+++ b/src/main/java/com/aetherteam/aether/event/listeners/RecipeListener.java
@@ -45,9 +45,10 @@ public class RecipeListener {
     @SubscribeEvent
     public static void onNeighborNotified(BlockEvent.NeighborNotifyEvent event) {
         LevelAccessor levelAccessor = event.getLevel();
-        if (!levelAccessor.dimensionType().effectsLocation().equals(AetherDimensions.AETHER_DIMENSION_TYPE.location()) return;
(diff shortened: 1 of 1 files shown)


bconlon1:
Icestone should still be able to work outside the Aether. I think isolating ban recipes to the Aether is potentially a good solution but i might have to think if theres other ways to avoid that overhead without possibly breaking changes.

shBLOCK:
Most of the overhead comes from biome matching for the ban recipe (for vanilla aether there are not many recipes to match for anyway).
So just adding this dimension check for the ban recipe is already going to get rid of most of the overhead.

bconlon1:
I may be able to fix some of the overhead by other means through caching inputs for more of the block recipes to check that list before getting to the biome check, I'll have to try it out.

bconlon1:
https://github.com/The-Aether-Team/The-Aether/pull/1959 wouldn't this be a better way of making this more performant?

shBLOCK:
> #1959 wouldn't this be a better way of making this more performant?

Oh yeah, I didn't even think of that, that's smart.
It would still cause a biome check if the block being updated is a torch or something, but I think one or two checks are probably fine?
But maybe it would still help to add this dimension check for the ban recipe.

bconlon1:
I would rather not as it would introduce a breaking change. I think this works for now unless people report continued performance issues.

Issue #1937, opened 2024-01-02
Bug: [1.20.1] Incompatibility with Stats Tinkerer "respawnHealth" setting


SigmaFlame:
Hello,

### What Feature Types Apply to This Bug?

Other (Please Describe)

### Other Type

Makes another mod lost a functionnality

### What Type of Bug Is This?

Compatibility

### Forge Version

47.2.19 (for minecraft 1.20.1)

### The Aether Version

1.0.0 (released 22/12/2023)

### Is This Bug a Conflict With Another Mod?

Stats Tinkerer

### Client Log

None

### Crash Report (if applicable)

None

### Steps to Reproduce

When configuring Stats Tinkerer to have only 6 HP after player death, try to die and respawn.

### What You Expect To Happen

Without your mod, the player revives with only 6 HP (3 hearth)

### What Actually Happened

With your mod, the player revives with all of their live.

### Additional Details

With mods that add life to the player, I really want them to respawn without all their hearts, because it makes the game too easy and players starts to die just to gain hearts and foods.

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

SigmaFlame:
https://github.com/Terrails/stat-tinkerer/issues/1

PR #1954, opened 2024-01-07
fix: check if respawn health is full

bconlon1:
#1937


OzPayn (approved):
seems to have fixed it
approved

Issue #1945, opened 2024-01-03
Enhancement: Check if world has a registered Aether level before allowing World Preview to open it.


bconlon1:
If the mod is installed on a world, the level for the Aether will always be registered to it. Therefore, I could determine if a world is unmodded with Aether by checking if its save contains the Aether level in some way. Probably.

PR #1955, opened 2024-01-07
fix: check if a world has aether before allowing world preview

bconlon1:
#1945


OzPayn (approved):
is working by preventing the world from being loaded, but I think maybe could do with some sort of UI element saying what is happening, right now it sort of just looks broken

functional though so approving, might open a separate issue for a UI addition at some point

PR #1958, opened 2024-01-07
fix: [Aerclouds] Removing "setOnGround" in entityInside method fixes undefined jumping behaviour

Trico-Everfire:
setOnGround is called the moment entityInside detects the player is inside the block, However, this causes undefined behaviour when the player jumps the moment they are inside the block, this is wrong as the player can basically jump on top of the block. This is unintentional and gets worse with the blue aether clouds where you can jump without triggering it's up effect.

With setOnGround:

https://github.com/The-Aether-Team/The-Aether/assets/55441008/36daf0b3-9f78-4963-88b6-1131c819999c

Without setOnGround:

https://github.com/The-Aether-Team/The-Aether/assets/55441008/fa2ccc69-6fcb-423e-9a8c-55b40eb5f213

The player regains the ability to jump once they've hit the bottom of the aercloud, which is inline with the original Aether.

### DIFF
--- a/src/main/java/com/aetherteam/aether/block/natural/BlueAercloudBlock.java
+++ b/src/main/java/com/aetherteam/aether/block/natural/BlueAercloudBlock.java
@@ -3,7 +3,6 @@ package com.aetherteam.aether.block.natural;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.particles.ParticleTypes;
 import net.minecraft.world.entity.Entity;
-import net.minecraft.world.entity.LivingEntity;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.entity.projectile.Projectile;
 import net.minecraft.world.level.BlockGetter;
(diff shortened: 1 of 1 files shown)


Trico-Everfire:
A weird thing I noticed is however that if the player jumps from a blue aercloud onto a cold/gold aercloud they do get to jump on top but only once, I have no idea what'd cause this as there is no logic for grounding them anymore.

bconlon1:
This actually _is_ intentional for cold aerclouds. However, it interfering with blue aerclouds' effect is not.

Trico-Everfire:
Really? It feels off from the original Aether. Having played that before I saw this project was released, it was quite weird to have you basically jump on top of a cloud like you do in this mod.

Trico-Everfire:
Okay, it seems like the code that's currently PRed is only for the Blue aerclouds, not white. So the PR is good to go afaik.

bconlon1 (approved):
lgtm

PR #1959, opened 2024-01-09
perf: reorder recipe biome checks

bconlon1:
#1949


shBLOCK (approved):
Clever and effective solution!

OzPayn (approved):
functionality seems to be retained, approved

Issue #1957, opened 2024-01-07
Feature Request: Add Swet Ball to forge:slimeballs


FinlaySML:
### Is This an Enhancement to a Preexisting Feature?

- [x] I confirm this request is an enhancement to a feature already in The Aether.

### What Is Your Feature Request?

Add swet balls to the forge "slimeballs" tag. I'm not sure if this would allow you to remove the swet ball recipes you've added for slimeblocks, sticky pistions, and leads.

### Why Should It Be in the Mod?

Because it allows for better integration with other mods recipes

### Please Read and Confirm the Following

- [X] I have confirmed this feature is either in the original Aether mod, or improves parity between The Aether and Minecraft or other mods.
- [X] I have confirmed this suggestion is unique and has not been suggested already.

PR #1960, opened 2024-01-09
feat: Add Swet Balls to Slimeballs tag

bconlon1:
#1957


OzPayn (approved):
cant see any change in functionality so I assume this is working correctly

PR #1963, opened 2024-01-13
fix: only hammer should allow custom lore

bconlon1:


OzPayn (approved):
think thats working
approved

PR #1969, opened 2024-01-15
fix: proper JER warning fix

### LINKED ISSUE #1967: Bug: Just Enough Resources conflict, MoaType is Null

Entity, Gui, Other (Please Describe)

### Other Type

mod conflict

### What Type of Bug Is This?

Compatibility

### Forge Version

47.2.19

### The Aether Version

1.1.0

Just Enough Resources

### Client Log

https://gist.github.com/katubug/94328eb27a666fdee3cd7014714b973b

_No response_

### Steps to Reproduce

Do /reload, check log

### What You Expect To Happen

I expect there not to be an error? lol

### What Actually Happened

Sorry, I know this is a weird report. I just found this line in my log while I was troubleshooting something else, and figured it would be good form to report it.

### Additional Details

_No response_

bconlon1:
#1967


Burning-Cactus (approved):
Looks good to me!

PR #1981, opened 2024-01-18
feat: 1.20.1 Implement REI Support

Dragon-Seeker:
PR adds support for REI(Roughly Enough Items). Such code was based on the existing JEI implementation with design changes based to reflect existing REI implementations. 

Minor fixes to the three primary menu textures to remove stray pixels present when using REI's dark mode.


bconlon1 (approved):
lgtm

PR #1993, opened 2024-01-21
feat: Add explicit support for Combatify

Alexandra-Myers:
After making the previous PR related to Combatify, I realized that the idea of only enabling abilities from The Aether if an attack is charged wouldn't work with my previous fix, as most attacks require 100% charge anyways when Combatify is installed. This means every ability would activate on every attack. Since this works against the design of The Aether's special abilities, I made this to ensure that those abilities only activate if an attack is "charged" by Combatify's standards. This is normally seen with sweeping or the charged reach bonus, both of which activate on 195% or above, hence why it is 195% or above here too. This won't impede on intended mechanics without Combatify, but will provide a more natural experience with it.

---

### DIFF
--- a/src/main/java/com/aetherteam/aether/item/EquipmentUtil.java
+++ b/src/main/java/com/aetherteam/aether/item/EquipmentUtil.java
@@ -9,6 +9,7 @@ import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.Item;
 import net.minecraft.world.item.ItemStack;
 import net.minecraftforge.common.util.LazyOptional;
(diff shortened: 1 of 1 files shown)


Alexandra-Myers:
I did an oopsie, apparently renaming the branch closes the pr

bconlon1 (approved):
lgtm

bconlon1:
As a contributor, would you be able to provide your player UUID?

Alexandra-Myers:
> As a contributor, would you be able to provide your player UUID?

b30c7223-3b1d-4099-ba1c-f4a45ba6e303

PR #2014, opened 2024-01-28
feat: jar size reduction

rhysdh540:
changes
 - compressed all png images with ect and oxipng
 - compressed all sounds with optivorbis
 - added a small script that minifies json files and re-zips the jar with java/zlib's highest compression setting
 - removed datagen `.cache/` folder from jar

should save at least 2-3 megabytes, probably more when combined with https://github.com/The-Aether-Team/Nitrogen/pull/21

I didn't realize that there was a commit naming policy, if necessary I can redo this, but squashing this pr should fix it anyways

---


bconlon1:
I can't check this out at the moment, but do block textures like leaves still have their transparent cutout parts made black when opaque with fast graphics? That has been an issue in the past with texture optimization attempts.

rhysdh540:
They do.
leaves
(the berry bush not becoming opaque also happens with the non-pr branch)

bconlon1:
Apologies for not having been able to get around to reviewing this yet. We've been in the middle of a porting period which has been taking longer than expected, so I plan to look more in-depth at this after we release the 1.20.4 port and focus on the 1.5.0 update milestone.

mrjasonn:
I suggest to please update the description to reflect that you are not using machete anymore, it will probably be easier for the Aether team. Thanks!

rhysdh540:
sorry! it should be more accurate now.

build.gradle
```diff
@@ -249,3 +257,41 @@ mixin {
 build {
     dependsOn 'jarJar'
 }
+
+import groovy.json.JsonOutput
```
mrjasonn: Would recommend to use gson instead for both parsing the json and outputting the json because it can handle comments in json files. I have tested using gson instead in my own mod that is a fabric mod and still minifies the jsons properly.

mrjasonn (changes requested):

build.gradle
rhysdh540: done!

mrjasonn (changes requested):
I would also suggest you can switch the branch to 1.20.4 and account for the updated textures that need to be compressed such as the updated berry bush texture

rhysdh540:
Should I do that by creating a new PR or by switching the branch on this one and merging everything?

mrjasonn:
> Should I do that by creating a new PR or by switching the branch on this one and merging everything?

Switching branch and merge everything then encompass new things and updated things etc (just would be my personal preference)

bconlon1:
Ended up implementing a different implementation of the first iteration of this as #2196. Keeping the image optimizations to the script instead of manually updating them on the repository is better for future-proofing, and the sound designers are more in charge of the audio optimizations. Using the machete plugin will also likely be better for maintainability from its simplicity.

Issue #2020, opened 2024-02-04
Feature Request: Tag for Gravitite toolset crafting items (and possibly other materials too for consistency)


Zepalesque:
### What Feature Types Apply to This Request?

Item

### Is This an Enhancement to a Preexisting Feature?

- [X] I confirm this request is an enhancement to a feature already in The Aether.

### What Is Your Feature Request?

A tag for items used to create Gravitite tools, weapons and armor (and possibly for consistency the other materials as well)
note: I can most likely (and probably will) make a pull request for this in case that’s helpful

### Why Should It Be in the Mod?

So that addons can change the materials used for different toolsets and keep compatibility with other addons (the specific use case I have in mind: I’d like to add Gravitite Ingots to my addon)

### Please Read and Confirm the Following

- [X] I have confirmed this feature is either in the original Aether mod, or improves parity between The Aether and Minecraft or other mods.
- [X] I have confirmed this suggestion is unique and has not been suggested already.

bconlon1:
Status:
- [x] 1.20.2
- [x] 1.20.1
- [x] 1.19.4
- [x] 1.19.2

PR #2023, opened 2024-02-04
feat(item/recipe): Tags for tool/armor set materials

Zepalesque:
This adds tags for armor set and tool set crafting materials. I didn't do Skyroot or Holystone since they're used in so many other things and are separate blocks, but I added ones for Gravitite and Zanite.

closes issue #2020 

---


bconlon1:
Going to do this on my end since I need to add some Nitrogen methods.

Issue #2009, opened 2024-01-26
Feature Request: Use a custom feature for bonemealing aether grass rather than `VegetationPlacements.GRASS_BONEMEAL`


Zepalesque:
### What Feature Types Apply to This Request?

Block, System

### Is This an Enhancement to a Preexisting Feature?

- [X] I confirm this request is an enhancement to a feature already in The Aether.

### What Is Your Feature Request?

Rather than using `VegetationPlacements.GRASS_BONEMEAL`, a custom grass feature (most likely a copy of the vanilla one) should be used [here,](https://github.com/The-Aether-Team/The-Aether/blob/e0a165ed64b3bb3801e402dd5a0707feac9983aa/src/main/java/com/aetherteam/aether/block/natural/AetherGrassBlock.java#L74) so that addons can replace the bonemealed grass without using mixins.

### Why Should It Be in the Mod?

So that addons don’t need to mixin into `AetherGrassBlock` or do weird stuff with events that could break in order to replace the grass spawned from bonemealing aether grass

### Please Read and Confirm the Following

- [X] I have confirmed this feature is either in the original Aether mod, or improves parity between The Aether and Minecraft or other mods.
- [X] I have confirmed this suggestion is unique and has not been suggested already.

PR #2024, opened 2024-02-05
refactor(block): Aether Grass now uses its own placed feature for bonemeal behavior

Zepalesque:
This pull request makes Aether Grass use a separate placed feature for when bone meal is used on it.

closes issue #2009 

not very relevant but this pull request's id is most likely going to be 2024 which I thought was funny since that's the current year

---


bconlon1 (approved):
lgtm

PR #2046, opened 2024-03-04
[1.20.1] refactor: Minor adjustment to AccessoryItem and variants

Dragon-Seeker:
- Add SoundEvent Supplier within base AccessoryItem
- Adjust constructors to call into overload variants for various Accessory implementation


bconlon1 (approved):
lgtm

PR #2087, opened 2024-04-02
improv: better dev mcfunctions

mrjasonn:
This pull request makes it so `dev_powerwash_chunk.mcfunction` uses `@p` instead of `Dev` as player (makes it consistent with other dev mcfunctions) and makes the unbreakability for gear in `suit_up.mcfunction` not use unbreaking 10 and instead use the unbreakable tag. `setup_structure_hunt.mcfunction` now uses infinite for the night vision instead of 9999 seconds as it seems like it was supposed to be infinite.
---

I agree to the Contributor License Agreement (CLA).

PR #2090, opened 2024-04-02
improv: make the night vision in dev mcfunction at least last as long as possible because no infinite on 1.19.2

mrjasonn:
**[THIS PR FOR 1.19.2 ONLY]** This PR makes the night vision at least last as long as possible (a million seconds) on 1.19.2 because infinite is only for 1.19.4 or higher. Apologies for not mentioning in #2087 that infinite is only for 1.19.4 or higher.

---


bconlon1 (approved):
lgtm

Issue #2100, opened 2024-04-04
Feature Request: Make Moa follow behavior partially controlled by Nature Staff


bconlon1:
Having shift right click activate following messes with being able to perform other interactions with that combination.

PR #2114, opened 2024-04-17
refactor: moa following depends on nature staff

bconlon1:
#2100


OzPayn (approved):
seems to be working! good stuff

Issue #2068, opened 2024-03-19
Bug: Aerbunny hops sometimes do not activate properly after gaining height


OzPayn:
In my experience holding the jump key while hitting a blue aercloud will generally not reset the aerbunny floating height properly, holding shift to sink into a blue aercloud before letting it launch you seems to make the behaviour more consistent/correct so it may be a hitbox issue with the aerclouds potentially?

PR #2120, opened 2024-04-18
fix: aerbunny jump reset bug

bconlon1:
#2068


OzPayn (approved):
Seems to be working reliably now, good job

Issue #2139, opened 2024-04-29
Enhancement: Moa behavior improvements


bconlon1:
- Vehicle checks should probably check for isSaddled as well.

PR #2144, opened 2024-05-03
refactor: moa checks isSaddled more

bconlon1:
#2139


OzPayn (approved):
this seems to be working

Issue #2130, opened 2024-04-24
Enhancement: Make portal activation more configurable


bconlon1:
remove config check from here so activation items and the config are handled independently https://github.com/The-Aether-Team/The-Aether/blob/9eb14e30531c39f9acded48baa85858b94180050/src/main/java/com/aetherteam/aether/event/hooks/DimensionHooks.java#L95

PR #2145, opened 2024-05-04
refactor: handle activation items and portal config separately

bconlon1:
#2130


OzPayn (approved):
was testing with a busted datapack (typo yay) but yes this appears to be working

PR #2152, opened 2024-05-05
improv: Mixin cleanup for compat and general improvements

Alexandra-Myers:
After perusing through the mixins to see what I could do (mixins tend to be a common source of issues, which I intended to solve), I noticed that none of them used MixinExtras features. 

While this isn't strictly required, and, indeed, should be used when it is proper practice to do so, there were several scenarios where it would be best practice, for mod compatibility and otherwise, to use it.

Hence, these changes are in regard to using MixinExtras in places where it should be.

---


src/main/java/com/aetherteam/aether/mixin/mixins/common/AbstractArrowMixin.java
```diff
@@ -29,25 +30,22 @@ private void tick(CallbackInfo ci) {
         AbstractArrow arrow = (AbstractArrow) (Object) this;
         if (arrow.hasData(AetherDataAttachments.PHOENIX_ARROW)) {
             var attachment = arrow.getData(AetherDataAttachments.PHOENIX_ARROW);
-            if (attachment.isPhoenixArrow()) {
-                if (!arrow.level().isClientSide()) {
-                    attachment.setSynched(arrow.getId(), INBTSynchable.Direction.CLIENT, "setPhoenixArrow", true); // Sync Phoenix Arrow variable to client.
-                    if (this.inGround) { // Spawn less particles when the arrow is in the ground.
-                        if (this.inGroundTime % 5 == 0) {
-                            for (int i = 0; i < 1; i++) {
```
Alexandra-Myers, line 37: This for loop will only ever run once, and can thus be replaced with just a single method call.

bconlon1 (changes requested):
Would you be able to split the accessory changes and the mixin changes into separate PRs?

Alexandra-Myers:
> Would you be able to split the accessory changes and the mixin changes into separate PRs?

Yes, admittedly I did it in one commit so it is rather difficult, but yes.

Alexandra-Myers:
Done.

Alexandra-Myers:
I realised that if this PR were to merge either before or after #2141, it would cause a merge conflict. I decided to revert FeatureMixin for this reason.

Alexandra-Myers:
?

Alexandra-Myers:
should be ready now

bconlon1 (approved):
lgtm

PR #2154, opened 2024-05-05
improv: Change accessories to hold their own accessory slot identifie…

Alexandra-Myers:
The way the specific accessories handled getting their slot identifiers involved plenty of identical code with minor variations depending on the item. This is resolved in the PR by introducing two things: 

1. An interface, which is used in cases where `instanceof` checks were used to identify identifiers, and provides one method, which provides the identifier it is trying to retrieve, and provides the benefit of needing fewer `if` statements to retrieve the same results.
2. A static method for each implementation of the interface to retrieve the identifier without needing to access an instance, but without requiring more repeated code. If, for whatever reason, the identifier was changed for an item, all usages, barring a few, would be changed just by changing this method.

---


bconlon1 (approved):
lgtm

Issue #2105, opened 2024-04-12
Feature Request: Config option to move Slider hit message (and maybe others) to action bar


Zepalesque:
### What Feature Types Apply to This Request?

Config, System

### Is This an Enhancement to a Preexisting Feature?

- [X] I confirm this request is an enhancement to a feature already in The Aether.

### What Is Your Feature Request?

A config option to move the slider message when attacking it with a non-pickaxe item to the status/action bar

### Why Should It Be in the Mod?

To be more consistent with things like beds, which no longer print their 'You can only sleep at night' and other messages directly to chat

### Please Read and Confirm the Following

- [X] I have confirmed this feature is either in the original Aether mod, or improves parity between The Aether and Minecraft or other mods.
- [X] I have confirmed this suggestion is unique and has not been suggested already.

PR #2174, opened 2024-05-29
feat: config for moving slider message

bconlon1:
#2105


OzPayn (approved):
that works (took me a while cos the way serverconfigs work has changed lmao)

Issue #2088, opened 2024-04-02
Enhancement: Valkyrie combat rebalance


OzPayn:
Investigate changes to slowfall for Valkyries to make them harder to cheese, maybe also some minor amount of knockback resistance would be appropriate across the valkyries generally would be appropriate (alongside potentially opening up ranged damage for the Queen that may make the fight more dynamic)

Potentially look into adjusting room layouts to make it harder for Valkyries to get stuck on everything too

OzPayn:
Valkyries should have more behaviour states between attacks, like repositioning on the ground or maybe even jumping into the air to try to evade?
I also think we should have a raise sword animation (like zombie arms) to signify when they're ready to attack (and there should be a small delay after they raise their arm before they can swing their sword) so blocking and avoiding attacks can be more reliable (shields are super unreliable as is so communicating stuff around it better would be great)

PR #2175, opened 2024-05-29
refactor: valkyrie balance tweak 

bconlon1:
#2088


OzPayn (approved):
these seem like positive changes so far, should help balance the dungeon a bit

Issue #2122, opened 2024-04-19
Bug: Lightning tracking edge cases


bconlon1:
There might still be an edge case allowing item destruction from thunder balls here if the projectile is dead already and can't be found in the world: https://github.com/The-Aether-Team/The-Aether/blob/c43520d1389727c2fddc5d08691a1e29ca2ed80a/src/main/java/com/aetherteam/aether/event/hooks/EntityHooks.java#L529

just make a non damaging lightning entity

PR #2176, opened 2024-05-29
fix: check correct thunder crystal owner

bconlon1:
#2122


OzPayn (approved):
seeeeems to be working

PR #2178, opened 2024-05-30
Fix Bronze Dungeon Vertical Tunnel crash

Drullkus:
Fixes crash when a Bronze Dungeon attempts to generate a tunnel shaft when its ceiling is higher than terrain's top surface. Also includes some other safety measures in case of wackiness from customizations


bconlon1 (approved):
lgtm

Issue #2035, opened 2024-02-24
Bug: Entities like Zephyrs still have some issues with spawning in spaces they shouldnt


bconlon1:
They seem to still be able to spawn in rooms and enclosed areas that aren't fully in view of skylight.

indrada:
I've been unable to reproduce this bug. Can you specify how this bug was produced or add screenshots of the location where the zephyr spawned?
![2024-06-09_01 02 35](https://github.com/The-Aether-Team/The-Aether/assets/61097641/1858ea34-24fd-49ba-8b7c-bcfd86ad7a03)

PR #2182, opened 2024-06-09
fix: Changed Aerwhale and Zephyr spawning behavior...

indrada:
fix: Changed Aerwhale and Zephyr spawning behavior to check entire hitbox for sky visibility

closes issue #2035

---


bconlon1 (changes requested):
It may be good to make a helper method in `aether.entity.EntityUtil` for this that takes the distance (like 1 or 2) as a parameter). The written code also doesn't seem to be used for the Aerwhale. For code cleanliness there should also always be spaces after commas, after `for`, and between `=` in local variable assignment.

bconlon1 (approved):

Issue #2181, opened 2024-06-04
[1.20.1] The Title Screen doesn't always switch to the Aether version


Flexico:
### What Feature Types Apply to This Bug?

Gui

### What Type of Bug Is This?

Annoyance

### Forge Version

47.2.32

### The Aether Version

1.4.2

### Client Log

https://gist.github.com/Flexico/292dca4160d01c5af696e242e11b0bcf

### Steps to Reproduce

Run the game; sometimes the Aether title screen appears, and sometimes it's just the normal Minecraft title screen. I can click the button to switch between them if the Aether version loads, but if the MC version shows up, the button is still there but nothing happens when I click it.

### What You Expect To Happen

Aether title screen appears

### What Actually Happened

About half the time I get the MC screen

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this bug is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this bug to their issue tracker already.
- [X] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

bconlon1:
Can you please provide your list of mods?

Flexico:
There's a bunch... ^^;
![image](https://github.com/The-Aether-Team/The-Aether/assets/5751770/88ec258f-01e1-488d-9cba-00f15bc4defd)

bconlon1:
In that case, would you be able to narrow down the list to see if you can find what which specific mod combination is causing the title screen issue?

Flexico:
Good call! I'll try that.
...
With only the Aether installed, I loaded the game 5 times and it showed the correct menu every time. I get a feeling the word "sometimes" will become a headache. XD
...
I'll get back with you when I have more experimental results.

PR #2184, opened 2024-06-10
fix: Adjusted toggleBetweenMenus function to never return null, in or…

indrada:
fix: Adjusted toggleBetweenMenus function to never return null, in order to fix menu button not working when other mods adjust active_menu

closes issue #2181


bconlon1 (approved):
lgtm

PR #2188, opened 2024-06-12
feat: Split Sentry jump and squish sound events, add a silent ambient sound with no subtitle for configurability purposes

Zepalesque:
This splits the squish and jump sound for Sentries, which previously both used the jump sound, and gives sentries a silent ambient sound without a subtitle, for the purpose of allowing it to be overriden with a resource pack ~~(such as in Redux's built-in packs lol)~~

---


src/main/java/com/aetherteam/aether/data/generators/AetherSoundData.java
```diff
@@ -558,6 +558,17 @@ public void registerSounds() {
...
+        this.add(AetherSoundEvents.ENTITY_SENTRY_SQUISH,
+                definition().with(
+                        sound("minecraft:mob/slime/big1"),
+                        sound("minecraft:mob/slime/big2"),
+                        sound("minecraft:mob/slime/big3"),
+                        sound("minecraft:mob/slime/big4")
+                ).subtitle("subtitles.aether.entity.sentry.squish")
+        );
+        this.add(AetherSoundEvents.ENTITY_SENTRY_AMBIENT,
+            definition().with(sound("aether:generic/silence"))
```
bconlon1: Instead of needing a silent file, couldn't you just not generate an entry for the unused ambient sound effect?

bconlon1 (changes requested):

src/main/java/com/aetherteam/aether/data/generators/AetherSoundData.java
Zepalesque: hmm, yeah good point

Zepalesque: ok, fixed it

bconlon1 (approved):
lgtm

PR #2195, opened 2024-06-18
fix: no more dark sky below y=60

bconlon1:


OzPayn (approved):
yep that seems to work

PR #2196, opened 2024-06-18
build: machete optimization

bconlon1:
based on #2014


OzPayn (approved):
lgtm

rhysdh540:
One note of warning—you’ve turned on debug symbol stripping for class files, and while that saves a lot of space, it also makes stacktraces with aether stack frames in them much harder to read. If that’s okay with you, then ignore this.

Issue #2115, opened 2024-04-17
Feature Request: Configurable dimension list for sun altar functionality


bconlon1:
Would need to abstract the time scale to not depend on the 3x length of day in the Aether.

PR #2200, opened 2024-06-21
feat: configurable sun altar dimension list

bconlon1:
#2115


OzPayn (approved):
seems to be working now!

Issue #1988, opened 2024-01-20
Feature Request: More Spawn Prevention


bconlon1:
Aechor Plants
- Placing flowers will prevent Aechor Plants from spawning in a range.

Swets
- Applying a Swet Cape to a banner will create a custom Swet Banner (made with vanilla patterns, similar to the Ominous Banner) which prevents Swet spawning.
- Swet Capes on armor stands and maybe possibly in item frames will also deter Swet spawning.

bconlon1:
Should probably include book of lore info for this.

PR #2201, opened 2024-06-21
feat: aechor plant and swet spawn prevention methods

bconlon1:
#1988


OzPayn (approved):
hard to test a negative but I think this is all working correctly now!

Issue #2128, opened 2024-04-23
Bug: Player relogs into the Aether even if logging out in another dimension


darklotus781:
### What Feature Types Apply to This Bug?

System

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

47.2.16

### The Aether Version

1.4.0

### Is This Bug a Conflict With Another Mod?

No

### Client Log

Nothing

### Steps to Reproduce

Occasionally when a player logs out in a different dimension, they will login in the Aether at the coords they logged out from which means the can essentially be trapped a long ways from their base or portal.
This seems to be related to the server setting in the Aether config to spawn in the Aether and not in the overworld.
It doesn't seem to be dimension specific either, can be Overworld -> Aether or The Nether -> Aether.

### What You Expect To Happen

Realistically, the player should relog and join the world in the dimension and coords they logged out from.

### What Actually Happened

Player logs in and is placed in the Aether at the coords they logged out from even if the dimension was different.

### Additional Details

I cannot reliably reproduce this but I wonder if it's a timing issue.  I haven't looked at your code to fully understand how it's determining when / where to place the player in the overworld but this is the only mod that does that, it's a modpack I built and is heavily customized with KubeJS but nothing that affects which dimension a player relogs or spawns in.

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

darklotus781:
FYI, this also affects players when they die.  Not all the time, but sometimes a player will die in the overworld or nether with a respawn anchor or bed in that dimension but still respawns in the Aether dimension instead of the other dimension.

PR #2203, opened 2024-06-22
improv: fix issues with spawning into aether

bconlon1:
#2128


OzPayn (approved):
struggled to replicate the bug but so far this doesn't seem to break anything so Im gonna approve it

PR #2214, opened 2024-06-25
feat: Add fields for Structure Processors to dungeon structures

Zepalesque:
This pull request adds Structure Processor fields to the dungeon structure codecs, with the purpose of them being so that addons can both add new blocks to the fields, as well as potentially modify the existing ones as well

---


bconlon1 (approved):
ltgm

Issue #2229, opened 2024-07-26
Bug: Flamarang (from Quark mod) does not damage Slider.


AxolotlAccount:
### What Feature Types Apply to This Bug?

Entity, Item

### What Type of Bug Is This?

Compatibility

### Forge Version

47.3.5

### The Aether Version

1.20.1-1.4.2-neoforge

### Client Log

https://gist.github.com/AxolotlAccount/d8ae2cc71205dd96621ab92c2a31a9be

### Steps to Reproduce

1. Use The Aether mod with Quark mod.
2. Enter Slider boss room.
3. Attack it by throwing Flamarang (from Quark mod) to it.

### What You Expect To Happen

Slider should get damaged, just like when Pickarang (from Quark mod) is thrown at it.

### What Actually Happened

Game gives "Hmm. Perhaps I need to attack it with a Pickaxe?" warning.

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this bug is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this bug to their issue tracker already.
- [X] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

PR #2240, opened 2024-08-16
fix: flamerang compat

bconlon1:
#2229


OzPayn (approved):
cant test but if this doesnt work when backported I will be angry
approved (but on thin ice)

Issue #2185, opened 2024-06-10
[1.20.x] Another Lootr Mimic Texture issue


noobanidus:
I had a mental note to "inform Artifacts/Aether developers" when I switched the texture atlases around and ... I didn't, and was only really reminded by bug reports and comments on CurseForge. Sorry!

### Forge Version

1.20.1

### The Aether Version

Any version linking Lootr pre-0.7.34.85

### Is This Bug a Conflict With Another Mod?

Lootr

### Steps to Reproduce

Install a newer version of Lootr (0.7.34.85, for example). Spawn a mimic!

### What You Expect To Happen

It to have the Lootr texture.

### What Actually Happened

It does not, because the developer of Lootr moved the textures for the chests from the `BLOCK_ATLAS` into the `CHEST_ATLAS` because it seemed like the *right thing* to do, only to cause compatibility issues with this mod and with Artifacts.

### Additional Details

Sorry!!!

I realize this now causes issues because if people have mismatched versions (an older version of Aether, newer version of Lootr OR, newer version of Aether, older version of Lootr) will cause some issues. I'm not even sure if there's a way to test for "does this texture exist? if not, use this one" within the contexts of entities.

The new atlas locations can now be found [here](https://github.com/LootrMinecraft/Lootr/blob/1.20/src/generated/resources/assets/minecraft/atlases/chests.json) if that helps!

Problematically, there's now a new texture system and configuration for Lootr. Unfortunately, it didn't occur to me to put it into the API (which is something that I've been trying to do), so if someone is using the "old_textures" or "vanilla_textures" option there isn't really a way to dynamically tell and adjust which textures you provide for the entity through the API...

(I promise I will edit this after it's posted because it's a bit chaotic with forms.)

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this bug is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this bug to their issue tracker already.
- [X] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

noobanidus:
Note: I haven't released this update for 1.20.4 Forge or Fabric yet! So ... there's more confusion.

Dr-WeiAL:
[...]

PR #2241, opened 2024-08-16
fix: lootr textures

bconlon1:
#2185


OzPayn:
should probably change the mimic texture to the new one as its the new default from what I can see
will do that tomorrow

OzPayn (approved):
lootr restored

Issue #2244, opened 2024-08-21
Feature Request: Crossbows Able to Be Repaired in Altar


PraseodymiumSpike:
### What Feature Types Apply to This Request?

Block, System

### Is This an Enhancement to a Preexisting Feature?

- [X] I confirm this request is an enhancement to a feature already in The Aether.

### What Is Your Feature Request?

Crossbows should be able to be repaired in the Altar like other Items with Durability. This should not delete Items loaded into the Crossbow like Fireworks and Arrows.

### Why Should It Be in the Mod?

It is consistent with other Items like Swords and even Bows.

### Please Read and Confirm the Following

- [X] I have confirmed this feature is either in the original Aether mod, or improves parity between The Aether and Minecraft or other mods.
- [X] I have confirmed this suggestion is unique and has not been suggested already.

PraseodymiumSpike:
Altars should also be able to repair Flint and Steel, and they are unable to.

PR #2251, opened 2024-08-29
improv: repairing recipes for crossbow and flint and steel; also reduced altar repair costs for vanilla items

bconlon1:
#2244


OzPayn (approved):
lgtm, new recipes work

Issue #2247, opened 2024-08-23
Bug: Crash due to AetherArmorMaterial


tobi1449:
### What Feature Types Apply to This Bug?

Item

### What Type of Bug Is This?

Crash

### Forge Version

47.3.5

### The Aether Version

1.4.2

### Is This Bug a Conflict With Another Mod?

Deeper Darker

### Crash Report (if applicable)

https://mclo.gs/JfdKI2q

### Steps to Reproduce

This happens quite often (first 1-3 times) when a new Apotheosis Boss is spawned via a gateway by Apotheotic Additions/Gateways to Eternity.

### What You Expect To Happen

No crash.

### What Actually Happened

Crashes

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this bug is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this bug to their issue tracker already.
- [X] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

PR #2252, opened 2024-08-29
fix: remove old OnlyIn annotation from AetherArmorMaterials#getName

bconlon1:
#2247


OzPayn (approved):
Gonna rely on wisdom of the elders on this one cos I have no idea what any of the dependant testing is or requires and just hope this works
tl;dr: lgtm

Issue #2230, opened 2024-07-31
Holystone Bricks has abnormally high blast resistance


BabylonAS:
### What Feature Types Apply to This Bug?

Block

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

20.4

### The Aether Version

1.4.1

### Steps to Reproduce

Explode something (such as TNT) on two surfaces: one composed of holystone bricks and other composed of End stone.

### What You Expect To Happen

The crater on the holystone brick surface must be larger, as expected of a material as weak as holystone.

### What Actually Happened

The crater on the holystone brick surface would be similar in size to the End stone one, maybe even smaller.

### Additional Details

End stone has a blast resistance of 9, whereas holystone bricks is at 10. At some point in Minecraft's development blast resistance values were downscaled by a factor of 5 (that is, End stone's blast resistance used to be 45). Holystone bricks' blast resistance was 10 in the old Aether versions for 1.2.5 and was never updated for the revised blast resistance scale, staying at 10 even in the Modding Legacy versions. The expected blast resistance for holystone bricks is 2, the same as vanilla bone blocks - higher than raw holystone but lower than similar artificial stone-type blocks.

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this bug is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this bug to their issue tracker already.
- [X] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

PR #2253, opened 2024-08-29
refactor: lower holystone brick blast resistance

bconlon1:
#2230


OzPayn (approved):
yep that looks consistent now

Issue #2098, opened 2024-04-03
Bug: Quark runic etching does not visually apply to gloves


Nicodemus111:
### What Feature Types Apply to This Bug?

Entity

### What Type of Bug Is This?

Compatibility

### Forge Version

1.20.1-47.1.3

### The Aether Version

1.3.1

### Is This Bug a Conflict With Another Mod?

Quark 1.20.1-4.0-439

### Steps to Reproduce

1. enchant gloves (any)
2. apply runic etching (any)
3. equip gloves (curio slot or aether slot does not matter)
4. default enchant glint

### What You Expect To Happen

it should show the runic etching

### What Actually Happened

it did not show the runic etching

### Additional Details

I had previously reported this to Quark:
[https://github.com/VazkiiMods/Quark/issues/4748](url)

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

PR #2254, opened 2024-08-29
fix: runic etchings now work on gloves

bconlon1:
#2098


OzPayn (approved):
runic etching works (pretty)

PR #2275, opened 2024-09-14
fix: item saving error with item attachment

### LINKED ISSUE #2228: Bug: the items completely disappear if you die a second time nearby without taking them. (NEOFORGE)

System

### Other Type

_No response_

### What Type of Bug Is This?

Annoyance

### Forge Version

20.4.237

### The Aether Version

1.20.4-1.4.1-neoforge

_No response_

### Client Log

https://gist.github.com/AbyssalGrasp/db8c615073617c4dd72e1f2e490111c7

_No response_

### Steps to Reproduce

This happens 100% of the time.
1. Put any items in your inventory.
2. Enter the other dimension (either Aether or Nether).
3. Go away from the portal for about 4-5 Chunks.
4. Die (or /kill)
5. Respawn, go into the portal again
6. Reach the death point.
7. DO NOT TAKE THE ITEMS and die again.
8. Respawn, go into the portal again
9. Reach the death point.
10. The items are completely disappeared.

### What You Expect To Happen

The items are expected to disapper afther 5 minutes have passed. If the chunk is unloaded, the timer shouldn't even start.

### What Actually Happened

Every time you die near the items that are left from your **previous** death, they would disappear, regardless of the time passing (one round of testing should only take around 40-50 seconds).

### Additional Details

The aether is THE ONLY mod. The problem occurs ONLY with Aether.
Basically, the bug happens when you die far enough (not necessairly in another dimension). All it needs is for the chunk you have died in to be unloaded.
The problem could have a chain effect, and it's especially harmful:
_You die and drop some items **(a)**_ ---> 
_You take some other items **(b)** to help you retrieve your original items **(a)**_ --->
_You go for your items **(a)**_. ---> 
_You've almost reached the items **(a)**, but something kills you and you drop your current items **(b)**_---> 
_You respawn, take some items **(c)** again_  --->  
_You reach the death point, but you see that items **(a)** have completely vanished_ ---> 
_At that moment you die again, without touching the items **(b)**_ --->
_You respawn, go to that death point and see that the items **(b)** have also vanished, the only ones left are items **(c)**._ 
In conclusion, you've lost 2 sets of items (**a** and **b**).
Everything took around 3 minutes, so they 100% have NOT despawned via normal minecraft means.

- [X] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this bug is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this bug to their issue tracker already.
- [X] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

bconlon1:
closes #2228

### DIFF
--- a/src/main/java/com/aetherteam/aether/attachment/DroppedItemAttachment.java
+++ b/src/main/java/com/aetherteam/aether/attachment/DroppedItemAttachment.java
@@ -9,6 +9,7 @@ import net.minecraft.world.level.Level;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.Collection;
+import java.util.Optional;
 
 /**
  * Capability class used to track {@link ItemEntity}s dropped by player death.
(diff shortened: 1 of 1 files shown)


OzPayn (approved):
fixed on my end gg

Issue #2277, opened 2024-09-16
Bug: My server crashes after a few minutes


Sk00ma:
### What Feature Types Apply to This Bug?

System

### What Type of Bug Is This?

Crash

### Forge Version

47.0.3

### The Aether Version

1.4.2

### Is This Bug a Conflict With Another Mod?

MCA Comes Alive Reborn

### Client Log

https://gist.github.com/Sk00ma/6e201686c57a217ba066cff178a51b52

### Crash Report (if applicable)

https://gist.github.com/Sk00ma/d8997dc8575dfa26233e1ed3c2390660

### Steps to Reproduce

This is happening every 10 minutes approximately since the server starts, it crashes and takes us all out of the server, this has been happening for a few days but it is the first time in a month using the mod inside a modpack that this happens

### What You Expect To Happen

I don't know what it can be, currently we have not made any changes in any mod and a week ago it was working perfectly

### What Actually Happened

I don't know what it can be, currently we have not made any changes in any mod and a week ago it was working perfectly

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this bug is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this bug to their issue tracker already.
- [X] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

bconlon1:
Seems to be caused by a specific entity which you might be able to kill with commands:

```Entity Type: mca:female_villager (forge.net.mca.entity.VillagerEntityMCA)
	Entity ID: 362234
	Entity Name: Johnny
	Entity's Exact location: 2014.37, 74.00, 3122.50
	Entity's Block location: World: (2014,74,3122), Section: (at 14,10,2 in 125,4,195; chunk contains blocks 2000,-64,3120 to 2015,319,3135), Region: (3,6; contains chunks 96,192 to 127,223, blocks 1536,-64,3072 to 2047,319,3583)
	Entity's Momentum: 0.01, -0.08, 0.00
	Entity's Passengers: []
	Entity's Vehicle: null

`````
I will have to look into the cause, might be the code assuming this entity should have attack damage when it doesn't.

Sk00ma:
[...]

PR #2278, opened 2024-09-17
fix: more secure check for vanilla equipment attributes

bconlon1:
#2277


OzPayn (approved):
lgtm dont fire me if stuff breaks

PR #2321, opened 2024-10-14
improv: use @s instead of @p in dev mcfunctions

mrjasonn:
We could have done this instead apparently.

---


oodrand:
Why? What's the purpose?

mrjasonn:
> Why? What's the purpose?

This would be a good change for development, of course

oodrand:
> This would be a good change for development, of course

No it wouldn't. You're nitpicking on small pieces of code just to add useless PRs.

PR #2326, opened 2024-10-26
Fixed crash when applying Gold Dungeon's SurfaceRuleProcessor to a modded biome

TunefulTurnip:
This pr resolves an issue regarding the SurfaceRuleProcessor used for the Gold Dungeon structure, when applied to a modded biome, due to the chunk not being present while generating the dungeon. This issue also applies if the Gold Dungeon is being generated in the overworld through a datapack. This issue is only present in version 1.21.+.


bconlon1 (approved):
lgtm

Issue #2306, opened 2024-10-07
Bug: Metallic rings, pendants, gloves and victory medals smelt in blast furnaces at the same speed as they do in regular furnaces, wasting fuel


BabylonAS:
### What Feature Types Apply to This Bug?

Item

### What Type of Bug Is This?

Unexpected Behavior

### Forge Version

20.4.237

### The Aether Version

1.5.0

### Steps to Reproduce

The items in question are: Iron Ring, Iron Pendant, Iron Gloves, Chainmail Gloves (smelt to Iron Nugget); Golden Ring, Golden Pendant, Golden Gloves, Victory Medal (smelt to Gold Nugget).
1. Place a regular furnace and a blast furnace.
2. Load both with fuel - ideally sticks, because they have (or supposed to have) a 1 to 1 smelting rate with the items in question, making it easier to see the issue.
3. In sequence, put two of each item into each furnace. See how quickly do the items smelt in each furnace.

### What You Expect To Happen

The blast furnace should smelt each item in question twice as fast as the regular furnace - just like it does with acceptable vanilla items (ores, raw ores, metallic tools, swords and armor). Both furnaces must consume the same amount of fuel.

### What Actually Happened

Both furnaces smelt the items in question at the same speed. However, the blast furnace still burns fuel at double rate, wasting half of it.

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this bug is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this bug to their issue tracker already.
- [X] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

PR #2331, opened 2024-10-30
fix: smelting speed

bconlon1:
#2306


OzPayn (approved):
smelt speed rejuvenated

BabylonAS:
Am I right to interpret the change as if the 2× smelting speed for the items affected by https://github.com/The-Aether-Team/The-Aether/issues/2306 was not intentional in the first place?

bconlon1:
Speed for smelting is 200 while blasting is 100

BabylonAS:
Yeah, that's how it's set for vanilla items. But when I first noticed the smelting speed quirk I thought the Aether deliberately had the “small” items like rings, necklaces, and medals smelt faster than all others. I assumed it was a feature, a neat usage of Minecraft's data pack based customizability, and the apparent bug was that the blasting speed wasn't increased in the same way. I didn't think it was the other way around.

The higher speed could make a bit less sense with gloves, but, on the other hand, a ring requires twice as many ingots. But, if the intent was to never have faster smelting for these items, then oh well.

PR #2336, opened 2024-11-03
fix: moa skin null check in moa renderers

bconlon1:


OzPayn (approved):
lgtm - those nulls sure are checked (I think)
moa skin functionality is not ruined so approving

Issue #2317, opened 2024-10-10
Bug: [1.19.2]Chest Mimic's texture got error when lootr is installed


FakeDaLaoQAQ:
### What Feature Types Apply to This Bug?

Block

### What Type of Bug Is This?

Compatibility

### Forge Version

43.3.9

### The Aether Version

1.5.0

### Is This Bug a Conflict With Another Mod?

Lootr

### Steps to Reproduce

1.Install Lootr to your game
2.join a world, put down a mimic_chest
3.got wrong texture

### What You Expect To Happen

Texture should be normal lootr's chest

### What Actually Happened

It turn Black And Purple

### Additional Details

Texture issue?

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this bug is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this bug to their issue tracker already.
- [X] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

FakeDaLaoQAQ:
![ERROR](https://github.com/user-attachments/assets/10aff990-d536-42b0-97c8-8af7a011e265)

PR #2338, opened 2024-11-04
fix: update lootr assets

bconlon1:
#2317


OzPayn (approved):
assets changed! very demure

Issue #2315, opened 2024-10-09
Feature Request(s): Toggleable random boss names


charleadev:
### What Feature Types Apply to This Request?

GUI and Items

### Is This an Enhancement to a Preexisting Feature?

- [x] I confirm this request is an enhancement to a feature already in The Aether.

### What Is Your Feature Request?

I made a PR to toggle the randomly-generated boss names, book of lore, and golden parachute for the 1.7.10 build of the mod > https://github.com/The-Aether-Team/The-Aether-Archived/pull/178

adding these features to modern versions of the mod would be a smart idea so these features aren't exclusive to a decade-old version of minecraft

### Why Should It Be in the Mod?

these changes mainly exist to benefit very-specific needs for modpack authors.

disabling the randomly generated boss names would be nice for modpack authors if their pack gives the bosses their own headcanon names, so instead of "(gibberish), (boss name)" it's just "(boss name)". the mob names themselves can be easily edited in a resource pack but as it is right now the random boss names are hard-coded in no matter what you do.

toggling the book of lore + golden parachute are also minor QoL changes for just in case you dont want them upon entering the aether for the first time. for example if you have a modpack with a lot of content in it then it doesn't make a whole lot of sense to have an item called "Book of Lore" that only works with Aether blocks and nothing else. Making their toggles separate like in the PR i provided would also be very nice for if someone wants to spawn with a golden parachute but not a book of lore and vice-versa.

### Please Read and Confirm the Following

- [X] I have confirmed this feature is either in the original Aether mod, or improves parity between The Aether and Minecraft or other mods.
- [X] I have confirmed this suggestion is unique and has not been suggested already.

PR #2339, opened 2024-11-04
feat: config for disabling randomized boss names

bconlon1:
#2315


OzPayn (approved):
great config for people who hate fun

looks good to me

PR #2340, opened 2024-11-05
fix: bronze room random selection bug

bconlon1:


OzPayn (approved):
this seems to be working so far

PR #2341, opened 2024-11-07
fix: catch NPE in perk verification

bconlon1:


OzPayn (approved):
so far so good hopefully this fixes it and doesnt cause more problems

PR #2371, opened 2024-11-13
Fix #2365

### LINKED ISSUE #2365: Bug: Silver Dungeons not spawning with clouds sometimes or always im not really sure

Drullkus:
Fixes Clouds not generating around Silver Dungeons, which is a non-deterministic worldgen bug caused by incomplete generation such as closing the save or structure being on edge of a passerby's chunk-loading.

There is also an additional measure to skip creation of Cloud structure pieces, if they will not place any blocks.

Fixes #2365


bconlon1 (approved):
lgtm

PR #2401, opened 2024-11-28
fix: adjust Projectile Impact event hook to only adjust event cancellation when invalid for hooking entities

Dragon-Seeker:


bconlon1 (approved):
lgtm

PR #2405, opened 2024-11-30
fix: Inability to sync if tools are debuffed to the player mining blocks

Dragon-Seeker:
Overall fixes issues where mining speed would be desynced between a client and server when debuffed tools were enabled under common conditions when on a server

---


bconlon1 (approved):
lgtm

Issue #1935, opened 2024-01-02
Bug: Full-height aerogel walls next to non-full-height aerogel walls do not have their faces culled as expected


muzikbike:
### What Feature Types Apply to This Bug?

Block

### What Type of Bug Is This?

Annoyance

### Forge Version

47.2.0

### The Aether Version

1.0.0-neoforge

### Steps to Reproduce

- Create a straight wall of aerogel wall blocks
- Place a single full block on top

### What You Expect To Happen

Internal planes would not be visible.

### What Actually Happened

They are where the wall elevation changes.

### Additional Details

Unsure as to if this is fixable with custom models alone. It may require the implementation of new block states specifically for this wall type as to select purpose-made model files.
![2023-12-28_03 21 15](https://github.com/The-Aether-Team/The-Aether/assets/52297970/02ccafc6-7669-4421-a476-6cd39768c27e)

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contained vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this issue is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this issue to their issue tracker already.

bconlon1:
Unfixable within Minecraft's model rendering system.

bconlon1:
Maybe actually fixable with some NeoForge systems.

PR #2473, opened 2025-01-02
fix: aerogel walls culling blocks they shouldn't

### LINKED ISSUE #1934: Bug: Aerogel walls should not cull the faces of adjacent opaque blocks

Block

### Other Type

_No response_

### What Type of Bug Is This?

Annoyance

### Forge Version

47.2.0

### The Aether Version

1.0.0-neoforge

_No response_

### Client Log

_No response_

_No response_

### Steps to Reproduce

- Place aerogel walls
- Place non-aerogel walls or a fance gate directly next to it so they connect directly

### What You Expect To Happen

The faces adjacent to the wall blocks would not be culled, since aerogel walls can be seen through.

### What Actually Happened

The faces are culled, which looks odd. Aerogel walls should only be able to cull the faces of other aerogel walls (but should still be culled by other walls).

embeddedt:
Fixes #1934. The strategy is to mark the blocks as `noOcclusion` (which I believe is more correct - they are translucent, so they won't ever be occluding geometry behind them) and then use `skipRendering` so they skip rendering their face when touching a wall of the exact same type.


embeddedt:
Hold on, just read #1935, and this fix changes the behavior of that in an interesting way. I'll see if I can address both issues at once.

bconlon1:
Yeah this was a tricky issue to try and figure out since the expected methods of solving it usually had other unforeseen side effects.

embeddedt:
Ok, I believe #1935 is not fixable without a custom baked model & extra logic (which is probably not worthwhile for one wall block), but it should behave the same way it did before this PR now.

bconlon1 (approved):
lgtm

Issue #2474, opened 2025-01-04
Bug: Double drops fail when a mob is killed without a tool


cassiancc:
### What Feature Types Apply to This Bug?

System

### What Type of Bug Is This?

Crash

### Mod Loader

Fabric

### Mod Loader Version

16.0.9

### The Aether Version

1.21.1-1.5.1-beta.4-fabric

### Is This Bug a Conflict With Another Mod?

Infinite Dimensions v2.3.8

### Client Log

https://mclo.gs/iGhxt2W

### Crash Report (if applicable)

https://mclo.gs/IGSm1m5

### Steps to Reproduce

1. Load a dimension with Chaos Slimes
2. Chaos Slimes die without player input, and thus no tool.
3. Aether tries to double the drop, but since there isn't a tool, the game crashes.

### What You Expect To Happen

The Chaos Slime dies without its drops being doubled, as it was not killed with a Skyroot item.

### What Actually Happened

The game/server crashes.

### Additional Details

Aether should not try and check what tool was used if the tool is null. The call to `toolStack.getItem()` in `DoubleDrops.java` fails as `toolStack` is null, and therefore does not have the method `getItem`

### Please Read and Confirm The Following

- [X] I have confirmed this bug can be replicated without the use of Optifine.
- [X] I have confirmed this bug is on the most recently supported version of Minecraft.
- [X] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [X] I have confirmed this bug is unique and has not been reported already.
- [X] If playing on a modpack, I have reported this bug to their issue tracker already.
- [X] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

bconlon1:
Can you test this with the full release of the fabric version and not the beta?

cassiancc:
Sure thing, but I had checked the code and the offending class/line hasn't been changed in three months.

bconlon1:
I think I know what the issue is but I'd still like to be sure

cassiancc:
Looks like it's been fixed in the released version. A null check in doubledrops might still be useful, but I haven't been able to get it to crash again in the latest update. Thanks for your help.

bconlon1:
Yeah I might just put one in anyways

cassiancc:
It's back (unfortunately)! Guess we might need that null check after all. 
https://mclo.gs/PyfejOj

PR #2475, opened 2025-01-04
fix: Null safety in DoubleDrops

cassiancc:
Adds basic null safety to double drops, as discussed in #2474. As the crash seems to be back in the released version, this is necessary.

closes issue #2474

---


bconlon1 (approved):
lgtm

PR #2496, opened 2025-01-17
WIP fix for portal client-server desync: type(feat): Incomplete but fixes dimension sync issue

Aquaeyes:
… by reusing base teleportation logic + hurtmarked

`(WIP fix for portal client-server desync)`: type(feat): Brief description *[MOVE TO TITLE]*

This doesn't work with passengers atm, and is putting players at weird y levels when 'falling' - there's an issue as well with Spawn Items being given multiple times, working on it

---


Aquaeyes:
Client desync where the client and server believe they're in different worlds should be fixed, alongside the player being 'blank' and the issue where block breaking can't occur

PR #2503, opened 2025-01-24
feat: config for disabling moa skins button; ignored for players with access to skins

bconlon1:
#2499


OzPayn (approved):
tested on accounts with skins and without, seems to be working as intended

PR #2586, opened 2025-02-18
fix: Curios Compat layer duplication bug

Dragon-Seeker:
Fixes issues with recent CCLayer changes to prevent duplication of accessories and makes the code only handle aether accessories for the data conversion process.

---


bconlon1 (approved):
lgtm

PR #2652, opened 2025-05-03
Feat:1.21.3 Port with Render Fix

baguchi:
This pr main to fix entity rendering such as EntityRenderState Stuff
also support AgeableMobRenderer


bconlon1 (approved):
lgtm

PR #2665, opened 2025-05-19
rendering and some block and entity behavior fix

baguchi:
- some Entity Behavior Error Fix(exept InterectResult)
- some Block Behavior Error fix
- and blit error fix


bconlon1 (approved):
lgtm

PR #2674, opened 2025-05-20
Networking update with N^2 complexity

KaiserOfNone:
`Networking update with N^2 complexity`: feat(networking)

Currently when a player joins any dimension, the server will launch N calls of `setShouldSyncBetweenClients`, once per each person connected to the server.
Once this happens each client will fire a call of `syncClients` which will fire a packet for each other player.
This sequence will saturate the network with many small calls, causing short but heavy lag spikes.

This fix was tested on a server with an average population of 20 and a max of 50 players.
Please tell me if there's any issues.

---


bconlon1 (approved):
lgtm

PR #2681, opened 2025-06-04
Fix onLoadPlayerFile method

Ultrasquid9:
The method should behave identically, it just has 8 less layers of nesting now 

---


bconlon1 (approved):
lgtm

PR #2682, opened 2025-06-04
Fix ClassCastException in drop modifiers

vaniron:
Fixes server crash I encountered [https://pastebin.com/raw/J1FSMUCW](https://pastebin.com/raw/J1FSMUCW) and also similar one in DoubleDropsModifier.


Dragon-Seeker:
@vaniron This fix dose not make sense as the `LootContextParam` are typed with `Entity` meaning anyone passing something that is not an entity there when making or adjusting the `LootContextParamSet` seems to be passing the wrong object.

Like this is an issue with a mod doing incorrect behavior and should not be fixed by us by making our code not error because someone is doing something wrong when using the API for loot context.

vaniron:
I see, its true i am not that much experienced with java so i had no idea whats the problem, just tried to come up with some solution for server not to crash. As I use it on production server with many players, i got no other option. I should better test mods more next time before adding them I guess.

bconlon1:
Yeah it sounds like this is some odd compatibility with another mod that you'll have to find.

Issue #2676, opened 2025-05-27
Bug: Aerbunny despawning (random)


SUPERNOAH66:
### What Feature Types Apply to This Bug?

Entity

### What Type of Bug Is This?

Unexpected Behavior

### Mod Loader

NeoForge

### Mod Loader Version

1.21.1

### The Aether Version

1.5.8

### Is This Bug a Conflict With Another Mod?

No

### Steps to Reproduce

Started playing the mod without any problem, To move between islands I would get an Aerbunny on my head and use the white tornadoes, no problem whatsoever, after I completed the bronze dungeon and equipped some stuff, when I’m traveling over the void with the Aerbunny (sometimes even on land) on my head it randomly despawns (seen in f5) and I’m not pressing shift or clicking the bunny/any kind of thing with the mouse, just pressing W and ctrl to run. It just despawns from the head and makes me fall, it started happening randomly and I don’t know why 

### What You Expect To Happen

Aerbunny sitting on my head

### What Actually Happened

Despawned randomly

### Please Read and Confirm The Following

- [x] I have confirmed this bug can be replicated without the use of Optifine.
- [x] I have confirmed this bug is on the most recently supported version of Minecraft.
- [x] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [x] I have confirmed this bug is unique and has not been reported already.
- [x] If playing on a modpack, I have reported this bug to their issue tracker already.
- [x] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

Issue #2664, opened 2025-05-19
Bug: <moas can't be riden in water>


catbryce:
### What Feature Types Apply to This Bug?

Entity

### What Type of Bug Is This?

Unexpected Behavior

### Mod Loader

forge

### Mod Loader Version

47.4.0

### The Aether Version

1.5.2

### Steps to Reproduce

whenever i try to cross a body of water or if i fall out of the aether into a body of water

### What You Expect To Happen

i expect to be able to ride moas on water they should be able to float like ducks
it works like that in the 1.7.10 version and the 1.2.5 version

### What Actually Happened

when i try to go across water on a moa is sinks until i get dismounted

### Please Read and Confirm The Following

- [x] I have confirmed this bug can be replicated without the use of Optifine.
- [x] I have confirmed this bug is on the most recently supported version of Minecraft.
- [x] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [x] I have confirmed this bug is unique and has not been reported already.
- [x] If playing on a modpack, I have reported this bug to their issue tracker already.
- [x] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

PR #2706, opened 2025-06-23
fix: aerbunny and moa and parachute fixes

### LINKED ISSUE #2632: Bug: Aerbunnies NOT preventing fall damage, and strange behavior when travelling through AE/OW portal

Entity

### Other Type

_No response_

### What Type of Bug Is This?

Unexpected Behavior

### Mod Loader

NeoForge

### Mod Loader Version

21.1.145

### The Aether Version

1.21.1-1.5.8

_No response_

### Client Log

_No response_

_No response_

### Steps to Reproduce

I could NOT reproduce the breakage it in a test world.

I theorize it is an interaction between having an Aerbunny on the player's head and something else like: Parachutes, bouncing on clouds, Zephyrs, Swet assaults.

I have tried: exiting and rejoining the world, reloading Minecraft, rebooting PC

Two Solutions I found:
1. Dying (not favorable)
2. Did not have an Aerbunny on my head in the Aether; traveled to Overworld and suddenly had an Aerbunny on my head; went back to AE and the bunny disappeared; went back to OW and it reappeared; took off bunny and sent it through the AE portal; went back to AE myself; saw bunny in my AE house and put on head; worked normally. (have not attempted to reproduce this fix, it was a strange one-time thing I happened upon).

### What You Expect To Happen

I expect to take no fall damage while Aerbunny is on top of head, and that the Aerbunny fall off player head where the player is.

### What Actually Happened

1. Aerbunny does NOT prevent fall damage from high heights
2. Aerbunny DOES still allow mid-air jumps
3. When right-clicked to take off Aerbunny from player head the Bunny remains suspended in air above player for a few seconds and then teleports back to where it was originally taken
4. Strange AE/OW portal behavior (see above)

### Additional Details

Parachutes deploy when they usually shouldn't when using an Aerbunny

- [x] I have confirmed this bug is on the most recently supported version of Minecraft.
- [x] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [x] I have confirmed this bug is unique and has not been reported already.
- [x] If playing on a modpack, I have reported this bug to their issue tracker already.
- [x] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

bconlon1:
#2632 
#2676 
#2664


Dragon-Seeker (approved):
Lgtm

PR #2714, opened 2025-06-28
refactor: change inebriation particle from red to purple

bconlon1:


OzPayn (approved):
no longer vomiting blood, good stuff

Issue #2717, opened 2025-07-01
Bug: Life Shard Hearts do not become hidden when F1 is pressed


UnknownSerhan:
### What Feature Types Apply to This Bug?

Other (Please Describe)

### Other Type

HUD issue

### What Type of Bug Is This?

Annoyance

### Mod Loader

NeoForge

### Mod Loader Version

1.21.1 - 21.1.176

### The Aether Version

1.5.8

### Steps to Reproduce

1. Use Life Shards.
2. Press F1.
3. See that the Life Shard hearts don't become hidden.

### What You Expect To Happen

The hearts are supposed to become hidden.

### What Actually Happened

The hearts don't become hidden. It is quite annoying while taking screenshots with F1.

### Additional Details

![Image](https://github.com/user-attachments/assets/64b67ecc-8496-438d-a6fa-146a28845e88)

I checked the other issues and couldn't find this issue. It is surprising that no one reported it yet.

### Please Read and Confirm The Following

- [x] I have confirmed this bug can be replicated without the use of Optifine.
- [x] I have confirmed this bug is on the most recently supported version of Minecraft.
- [x] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [x] I have confirmed this bug is unique and has not been reported already.
- [x] If playing on a modpack, I have reported this bug to their issue tracker already.
- [x] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

PR #2722, opened 2025-07-08
Fix: AetherOverlays not affected by hideGui value

darksonic300:
Fixes various overlays that were still rendering even after hiding the player HUD.
Simply adds value checking.

closes issue #2717

---


bconlon1 (approved):
lgtm

PR #2738, opened 2025-08-01
fix: use IClientItemExtensions for default glove color

GizmoTheMoonPig:
Fixes gloves not using the color defined in the IClientItemExtensions, instead hardcoding it to the leather color.

Umbral Skies adds gloves for TF's Arctic armor, where we change the default color to better match the armor. I would honestly like to change the glove rendering system a bit to support multiple layers on gloves using the `ArmorMaterial.Layer` stuff but that can wait until I have a bit more free time. 


bconlon1 (approved):
lgtm

PR #2748, opened 2025-08-06
feat: 1.5.9 update work

bconlon1:
#2627 
#2569 
#2515
#2745
#2772
#2749

Issue #2569, opened 2025-02-14
Bug: Milk Bucket Fluid Capabilities


VoidLeech:
### What Feature Types Apply to This Bug?

Item

### What Type of Bug Is This?

Unexpected Behavior

### Mod Loader

Neoforge

### Mod Loader Version

47.1.70, 21.1.72

### The Aether Version

1.5.2, 1.5.4

### Steps to Reproduce

1. Query for ForgeCapabilities.FLUID_HANDLER_ITEM/Capabilities.FluidHandler.ITEM on a stack of the Skyroot Milk Bucket
2. Call getContainer() on the handler.

### What You Expect To Happen

The returned container is a skyroot bucket.

### What Actually Happened

On 1.21.1, there is no capability. 
On 1.20.1, this returned container is a normal bucket, not a skyroot bucket. This can seen in action with a mod like Create, which generates certain recipes from fluid capabilities: https://github.com/Creators-of-Create/Create/pull/7084

### Additional Details

Code analysis:
1.20.1 registered the capability via MilkBucketItem while 1.21.1 doesn't, explaining the missing capability.

1.20.1's MilkBucketItem's initCapabilities should be overridden: it currently returns a FluidBucketWrapper, whose default setFluid method will set the container to a normal bucket if the fluid is empty)

### Please Read and Confirm The Following

- [x] I have confirmed this bug can be replicated without the use of Optifine.
- [x] I have confirmed this bug is on the most recently supported version of Minecraft.
- [x] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [x] I have confirmed this bug is unique and has not been reported already.
- [x] If playing on a modpack, I have reported this bug to their issue tracker already.
- [x] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

PR #2753, opened 2025-08-07
fix: capability not allowing filling by milk

### LINKED ISSUE #2752: Bug: skyroot bucket

Item

### Other Type

_No response_

### What Type of Bug Is This?

Compatibility

### Mod Loader

NeoForge

### Mod Loader Version

39.0.0

### The Aether Version

1.1.1

Create

### Client Log

_No response_

_No response_

### Steps to Reproduce

Make this set up

Image

### What You Expect To Happen

Spout will dispense milk to bucket

### What Actually Happened

spout will be put into a loop of attempting and failing to put milk into bucket

### Additional Details

_No response_

- [x] I have confirmed this bug is on the most recently supported version of Minecraft.
- [x] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [x] I have confirmed this bug is unique and has not been reported already.
- [x] If playing on a modpack, I have reported this bug to their issue tracker already.
- [x] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

VoidLeech:
Milk wasn't allowed as fluid for filling SkyrootBucketWrapper, and then afterwards was not actually in the REPLACEMENTS map to actually be swapped.

Fixes #2752

Relates to #2569 so would be nice for this to be included in #2748 as well.

---


bconlon1 (approved):
lgtm

Issue #2785, opened 2025-09-24
Bug: <title>


Aereon767455:
### What Feature Types Apply to This Bug?

Item

### What Type of Bug Is This?

Unexpected Behavior

### Mod Loader

NeoForge

### Mod Loader Version

21.1.200

### The Aether Version

1.5.9

### Steps to Reproduce

1. Attempt to apply Mending, Unbreaking, or other Durability based enchantments to a Cloud Staff using an Enchanted Book on an Anvil or other equivalent enchanting method.
2. Attempt to apply Sweeping Edge to a Cloud Staff using an Enchanted Book on an Anvil or other equivalent enchanting method.

### What You Expect To Happen

The Cloud Staff should be compatible with Mending, Unbreaking, or other Durability based enchantments.
The Valkyrie Lance should not be compatible with Sweeping Edge.

### What Actually Happened

The Cloud Staff is incompatible with Mending, Unbreaking, and other Durability enchantments.
The Valkyrie Lance is compatible with Sweeping Edge.

### Additional Details

I was told by bconlon on the Discord for The Aether Project to report this on the GitHub, and that those might have gotten messed up when enchantments changed in 1.21.

### Please Read and Confirm The Following

- [x] I have confirmed this bug can be replicated without the use of Optifine.
- [x] I have confirmed this bug is on the most recently supported version of Minecraft.
- [x] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [x] I have confirmed this bug is unique and has not been reported already.
- [x] If playing on a modpack, I have reported this bug to their issue tracker already.
- [x] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

PR #2787, opened 2025-09-24
fix: cloud staff, valkyrie lance, and dart shooter enchantments

bconlon1:
#2785


baguchi:
looks like need to put ItemTags.DURABILITY_ENCHANTABLE
because ItemTags.BOW_ENCHANTABLE is not include durability things
so that's why unbreaking and mending cannot put dart shooter

baguchi (approved):
It worked! good job ;)

Issue #2798, opened 2025-10-03
Feature Request: Rain should kill Swets


Qwerty97475:
### What Feature Types Apply to This Request?

System

### Is This an Enhancement to a Preexisting Feature?

- [x] I confirm this request is an enhancement to a feature already in The Aether.

### What Is Your Feature Request?

Rain should kill Swets, because Swets shrink/die in water and rain is made of water

### Why Should It Be in the Mod?

Makes sense, as Endermen take damage from water and rain, so Swets should follow a similar logic.

### Please Read and Confirm the Following

- [x] I have confirmed this feature is either in the original Aether mod, or improves parity between The Aether and Minecraft or other mods.
- [x] I have confirmed this suggestion is unique and has not been suggested already.

PR #2800, opened 2025-10-04
refactor: swets check rain and bubble columns to dissolve

bconlon1:
#2798


OzPayn (approved):
seems to work correctly

Issue #2809, opened 2025-10-08
Bug: Large amounts of Evil Whirlwinds create crazy (TPS?) lag in the Nether


Qwerty97475:
### What Feature Types Apply to This Bug?

Entity

### What Type of Bug Is This?

Performance

### Mod Loader

NeoForge

### Mod Loader Version

21.1.209

### The Aether Version

1.5.10

### Client Log

https://gist.github.com/Qwerty97475/8c3358ceb30f780b57757e15d7ba7747

### Steps to Reproduce

1. Load up game with my modlist
2. Create new superflat creative world
3. Go to the Nether and dig a ditch in the ground to spawn Evil Whirlwinds in
3. Spam a bunch of Evil Whirlwind spawn eggs in the ditch
4. Observe lag, as messages take like 10s to register, dropping items doesn't create a dropped item entity, etc

<img width="1920" height="991" alt="Image" src="https://github.com/user-attachments/assets/4c7e1607-dc64-4c13-86ae-3b26817606ea" />

### What You Expect To Happen

Evil Whirlwinds wouldn't be so laggy

### What Actually Happened

Evil Whirlwinds were laggy

### Please Read and Confirm The Following

- [x] I have confirmed this bug can be replicated without the use of Optifine.
- [x] I have confirmed this bug is on the most recently supported version of Minecraft.
- [x] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [x] I have confirmed this bug is unique and has not been reported already.
- [x] If playing on a modpack, I have reported this bug to their issue tracker already.
- [x] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

bconlon1:
Likely related to #2782

Issue #2782, opened 2025-09-23
Bug: Whirlwind's pathfinding can cause server to hang and crash


staridiot:
### What Feature Types Apply to This Bug?

Entity

### What Type of Bug Is This?

Crash

### Mod Loader

NeoForge

### Mod Loader Version

21.1.209

### The Aether Version

1.5.9

### Is This Bug a Conflict With Another Mod?

Only Aether and its dependency owo present

### Crash Report (if applicable)

https://gist.github.com/staridiot/01f48b0f615aa53b04bd8a8f40c7eed0

### Steps to Reproduce

Reproducing this crash is quite tricky, honestly not sure how it happens natural.

I successfully reproduced this once by placing a top-slab, surround the slab in solid blocks and water-logging the slab. I'm pretty sure whats important here is the height, these 5 blocks are the only connected blocks at this height.

I then spawned 20-30 whirlwinds with the whirlwind spawn egg on the waterlogged slab. This caused the server to hang and crash, if your client disconnects before the server closes it may recover since the world unloaded.

### What You Expect To Happen

Obviously something wrong with the whirlwind path finding, 30-40 whirlwinds shouldn't hang the server. I also imagine this happened with a lot less when this happened naturally.

### What Actually Happened

The server hangs and eventually crashes

### Additional Details

I initially experienced this issue on the Allthemods 10 modpack. However I confirmed this issue still occurs with just The Aether and its dependency.

### Please Read and Confirm The Following

- [x] I have confirmed this bug can be replicated without the use of Optifine.
- [x] I have confirmed this bug is on the most recently supported version of Minecraft.
- [x] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [x] I have confirmed this bug is unique and has not been reported already.
- [x] If playing on a modpack, I have reported this bug to their issue tracker already.
- [x] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

PR #2810, opened 2025-10-08
fix: whirlwinds shouldnt be affected by other whirlwinds

bconlon1:
#2782 
#2809


baguchi (approved):
looks like good for me!

PR #2819, opened 2025-10-31
fix: remove Combatify support explicitly.

Alexandra-Myers:
Currently, the Combatify support implemented in #1993, works under its default config, however since then, there is now a config option for whether attacks should charge up to 200%. This leads to the abilities of Aether weapons not working if the user turns off charged attacks, hence this PR reverts the changes, which will be reimplemented separately on Combatify's end.

---


Alexandra-Myers:
I had preferred not adding it as a compile-only library, but there's not a realistic option since it would require reflection to be able to access the necessary method and the config itself, and I would rather not add reflection if it were unnecessary. Just reading the config from the file is also not an option due to how Atlas Core handles config options.

Alexandra-Myers:
On second thought, this won't compile without Atlas Core present, so perhaps reflection is better?

Alexandra-Myers:
I have decided to instead simply remove Combatify support directly, instead I will probably implement it on Combatify's end or as a compat mod.

Issue #2770, opened 2025-09-07
Bug: Stuck on "Joining world..." until server restart


marleyinbloom:
### What Feature Types Apply to This Bug?

Entity

### What Type of Bug Is This?

Performance

### Mod Loader

NeoForge

### Mod Loader Version

21.1.201

### The Aether Version

1.5.9

### Client Log

https://gist.github.com/marleyinbloom/cd854ab14a992a49d073ed933bc50b29

### Steps to Reproduce

Bug seems to be related to Moas in some way but I'm uncertain of the specifics, I haven't done much in the Aether, but my players have. There is a large pen of baby Moas near the world spawn that may have something to do with it?

### What You Expect To Happen

I'd expect to be able to join the server without much issue

### What Actually Happened

I get stuck on the "Loading world..." screen, an error appears in the server console but nothing of note on the client end.

### Additional Details

On the server end, upon joining, this error appears:
```
[01:15:53] [User Authenticator #24/INFO] [minecraft/ServerLoginPacketListenerImpl]: UUID of player marleyinblooom is 2a96f756-a39e-4720-ae20-d9f2812c5046
[01:15:53] [Server thread/ERROR] [minecraft/ServerPacketListener]: Failed to handle packet ServerboundSelectKnownPacks[knownPacks=[minecraft:core:1.21.1]], suppressing error
java.lang.IllegalArgumentException: Failed to serialize ResourceKey[aether:moa_type / aether:black]: Item must not be minecraft:air
        at TRANSFORMER/minecraft@1.21.1/net.minecraft.core.RegistrySynchronization.lambda$packRegistry$1(RegistrySynchronization.java:39) ~[server-1.21.1-20240808.144430-srg.jar%23200!/:?] {re:classloading}
        at MC-BOOTSTRAP/datafixerupper@8.0.16/com.mojang.serialization.DataResult$Error.getOrThrow(DataResult.java:287) ~[datafixerupper-8.0.16.jar%23100!/:?] {re:mixin}
        at TRANSFORMER/minecraft@1.21.1/net.minecraft.core.RegistrySynchronization.lambda$packRegistry$2(RegistrySynchronization.java:39) ~[server-1.21.1-20240808.144430-srg.jar%23200!/:?] {re:classloading}
[...]

PR #2828, opened 2025-12-21
Possible fix for issue #2770.

darksonic300:
Based on errors of various users, and a brief code analysis, the cause of issue #2770 could be related to the MoaType CODEC retrieval somehow failing to read the egg item, as if the item didn't exist.

I changed the codec definition to be lazy initialized, with a fallback Vanilla egg, so that players should supposedly still be able to join servers that find themselves in such a situation.

I couldn't test the fix as I cannot recreate the conditions of the issue. This means it might not be enough to resolve it.

---


Mats1300:
Did this Fix the issue?

Issue #2816, opened 2025-10-23
Bug: NoSuchMethodError related to aetherFabric$holderOrThrow


Krynnx3D:
### What Feature Types Apply to This Bug?

Entity

### What Type of Bug Is This?

Crash

### Mod Loader

Fabric

### Mod Loader Version

fabric-api-0.116.7+1.21.1

### The Aether Version

aether-1.21.1-1.5.11-fabric

### Is This Bug a Conflict With Another Mod?

Accessories (1.1.0-beta.52)

### Crash Report (if applicable)

https://gist.github.com/Krynnx3D/fc8e63e623686ebc338ca2800256736d

### Steps to Reproduce

This happens when a player dies in any ways

### What You Expect To Happen

Player dies normally

### What Actually Happened

Server Crashes

### Please Read and Confirm The Following

- [x] I have confirmed this bug can be replicated without the use of Optifine.
- [x] I have confirmed this bug is on the most recently supported version of Minecraft.
- [x] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [x] I have confirmed this bug is unique and has not been reported already.
- [x] If playing on a modpack, I have reported this bug to their issue tracker already.
- [x] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

thyTwilightGoth:
I can confirm that I have the same crash
[https://gist.github.com/thyTwilightGoth/d88440c0e8cbd3e2d89f7934a9197f68](https://gist.github.com/thyTwilightGoth/d88440c0e8cbd3e2d89f7934a9197f68)

firefloc:
Update: fix is available upstream on the Fabric 1.21.1 mainline.

Fix reference (upstream source):
https://github.com/The-Aether-Team/The-Aether/blob/1.21.1-develop-fabric/src/main/java/com/aetherteam/aether/event/listeners/EntityListener.java#L55

The looting lookup path in `EntityListener` is handled directly in upstream code.

Issue #2815, opened 2025-10-16
Bug: Dedicated Server Crash: NoSuchMethodError on Player Login


Lopagg:
### What Feature Types Apply to This Bug?

Entity

### What Type of Bug Is This?

Crash

### Mod Loader

Fabric

### Mod Loader Version

0.17.3

### The Aether Version

1.21.1-1.5.11

### Is This Bug a Conflict With Another Mod?

Accessories apparently, even though it's not installed separately

### Client Log

https://gist.github.com/Lopagg/6ac4b9e10cab49f998c4b2c152306c75

### Crash Report (if applicable)

https://gist.github.com/Lopagg/21d361cb1f5f994278073961eb70d4d7

### Steps to Reproduce

1.  Set up a  Fabric 1.21.1 dedicated server
2.  Install The Aether (`1.5.11`)
3. Start the server
4. Join the server as a player
5. Wait between a few seconds and a few minutes
6. The server crashes

Apparently the crash happens after an entity dies.

### What You Expect To Happen

I expected the server not to crash

### What Actually Happened

The server crashes

### Additional Details

The crash only happens on the server, not in single player worlds.
In the server world I had previously installed trinkets. I then tried to install Accessories and Accessories Compatibility Layer, but the issue is the same.

### Please Read and Confirm The Following

- [x] I have confirmed this bug can be replicated without the use of Optifine.
- [x] I have confirmed this bug is on the most recently supported version of Minecraft.
- [x] I have confirmed the details provided in this report are concise as possible and does not contain vague information (ie. Versions are properly recorded, answers to questions are clear).
- [x] I have confirmed this bug is unique and has not been reported already.
- [x] If playing on a modpack, I have reported this bug to their issue tracker already.
- [x] I have confirmed that I'm reporting a bug in The Aether I, not The Aether II.

firefloc:
Update: fix is available upstream on the Fabric 1.21.1 mainline.

Fix reference (upstream source):
https://github.com/The-Aether-Team/The-Aether/blob/1.21.1-develop-fabric/src/main/java/com/aetherteam/aether/event/listeners/EntityListener.java#L55

The looting lookup path in `EntityListener` is handled directly in upstream code.

PR #2852, opened 2026-02-22
Fix dedicated-server crash by avoiding mixin-injected holder lookup in death callback

firefloc:
Summary
This PR fixes a dedicated-server crash path in `EntityListener` by removing the call to `registryAccess().aetherFabric$holderOrThrow(...)` and using vanilla holder lookup directly.

Problem
On dedicated servers, the death callback can hit:
- `java.lang.NoSuchMethodError: DynamicRegistryManager.aetherFabric$holderOrThrow(...)`

This happens in the loot calculation path for Accessories `OnDeathCallback` when getting the Looting enchantment holder.

Change
File:
- `src/main/java/com/aetherteam/aether/event/listeners/EntityListener.java`

Replacement:
- before: `registryAccess().aetherFabric$holderOrThrow(Enchantments.LOOTING)`
- after: `registryAccess().lookupOrThrow(Enchantments.LOOTING.registryKey()).getOrThrow(Enchantments.LOOTING)`

Why this is safer
- Uses vanilla registry access APIs only.
- Removes dependency on a mixin-injected extension method in a critical server death path.
- Keeps behavior identical for Looting level computation.

Validation performed
- Change is isolated to one file and one call path.
- Local Gradle compile could not be completed in this environment due Java/Gradle compatibility (`Unsupported class file major version 69`), so CI validation is requested.

Issues
- Intended to address dedicated-server crash reports equivalent to #2815 / #2816.

Disclosure
- This fix was vibe coded with assistance from LLM tooling, then manually reviewed and minimized to a single targeted change.

### DIFF
--- a/src/main/java/com/aetherteam/aether/event/listeners/EntityListener.java
+++ b/src/main/java/com/aetherteam/aether/event/listeners/EntityListener.java
@@ -52,7 +52,10 @@ public class EntityListener {
         OnDeathCallback.EVENT.register((currentState, entity, capability, damageSource, droppedStacks) -> {
             List<ItemStack> droppedStacksCopy = new ArrayList<>(droppedStacks);
             boolean recentlyHit = entity.hurtMarked;
-            int looting = EnchantmentHelper.getEnchantmentLevel(entity.level().registryAccess().aetherFabric$holderOrThrow(Enchantments.LOOTING), entity);
+            int looting = EnchantmentHelper.getEnchantmentLevel(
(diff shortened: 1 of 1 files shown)


firefloc:
Closing this PR as superseded by the upstream fix already present in the Fabric 1.21.1 mainline branch.

Relevant upstream fix location:
https://github.com/The-Aether-Team/The-Aether/blob/1.21.1-develop-fabric/src/main/java/com/aetherteam/aether/event/listeners/EntityListener.java#L55

This addresses the dedicated-server looting holder lookup path directly in upstream source.

